Issue #3, opened 2017-11-12
Breaks when upgrading Gradle to 4.x


lnhrdt:
I've been using kotlinconf-app as a reference for how to get a Kotlin React application to work with a Kotlin Multiplatform dependency. Super cool stuff!

I noticed that the Gradle wrapper in this project is set to v3.3 and the setup here only works with v3.x. When I try to upgrade to a more recent v4.x Gradle version I get these Webpack errors:

```
Hash: da24f46d6a2df326d929
Version: webpack 3.4.1
Time: 159ms
        Asset       Size  Chunks             Chunk Names
    bundle.js  719 bytes       0  [emitted]  main
bundle.js.map    6.42 kB       0  [emitted]  main
   [0] ./src/main/js/index.js 43 bytes {0} [built]

ERROR in ./src/main/js/index.js
Module not found: Error: Can't resolve 'common-js.js' in '/Users/leonhardt/workspace/kotlinconf-app/web/src/main/js'
 @ ./src/main/js/index.js 1:0-23

ERROR in ./src/main/js/index.js
Module not found: Error: Can't resolve 'web.js' in '/Users/leonhardt/workspace/kotlinconf-app/web/src/main/js'
 @ ./src/main/js/index.js 2:0-17
```

I'd like to get this working with Gradle v4.x because I'm trying to integrate this approach into an existing Gradle multi-project application that has a Spring Boot application. The `org.springframework.boot` Gradle plugin requires Gradle 4.0 or later (https://github.com/spring-projects/spring-boot/issues/9516).

Any ideas of how to get a Gradle setup like the one here in kotlinconf-app to play nice with Gradle v4.x or even just what's going wrong when I try it?

jarekankowski:
@lnhrdt Thank you for raising this. Gradle plugin has been updated to `4.9-rc-1` with https://github.com/JetBrains/kotlinconf-app/pull/14.

PR #6, opened 2017-12-01
upgrade to gradle wrapper v4.2

fboldog:
This PR upgrade and fix:
- gradle wrapper to v4.2
- android gradle plugin to 3.0.1
- fixes build problem in the web project (#3)


yole:
We intentionally use an old Gradle and Android plugin version because the new Gradle plugin requires Android Studio 3, and Android Studio has no support for running JS tests. It may be possible to update the version once IntelliJ IDEA 2018.1 is out.

fboldog:
Good to know, thanks!

PR #11, opened 2018-07-17
Removed the web and common-js modules as they are not needed anymore

jamescoggan:
### Description

Removing the `web` and `common-js` modules. They aren't needed anymore.


yole:
What do you mean by "not needed anymore"? We're using kotlinconf-app as one of the official demo projects for Kotlin multiplatform projects, and these modules are part of this demo.

jamescoggan:
It was agreed with Hadi, but your comment make's sense. I'll make sure to keep it!
Thanks!

PR #17, opened 2018-08-22
[WIP] Use ktor mpp-client

e5l:
WIP, but we could start the review.

Serialization is missing for now.


MarcinMoskala:
Does it work without Serialization?

e5l:
Android part is fully working.

PR #18, opened 2018-08-22
Backend code verification and sessionize sync trigger

MarcinMoskala:
As discussed, this PR introduces 2 endpoints:

GET /users/verify/{code} 
returns code 200 (if code exists) or 406 (if code is not correct) 
Used when a user will be introducing a code to the application, to say if the code is correct or not. 

GET /sessionizeSync
Used to trigger synchronisation with Sessionize. It is needed because we changed refresh time to 1h from 5 minutes. This endpoint should be used after some important data were changed in Sessionize. 

Additional comment: We already have a mechanism for user identification. Right now Android app generates Android ID by `"android-" + UUID.randomUUID().toString()` and saves it to shared preferences using "UserId" key (iOS works the same way). Then it is posted to endpoint "POST /users". This adds this id to the database of users. From this point, every request has in header "Authorization" field with value "Bearer $userId" and all person-specific requests identify person first. Also, table with votes holds info about the user so when user votes again, his vote is changed instead of added. 

This mechanism is good. The only difference we should do is that instead of generating and registering code, the user should introduce his code into the app. Once this mechanism on applications will be ready, I will block endpoint for users creation (I will restrict it with some secret). At the moment we have in the backend everything we need for new systems of codes.


backend/src/org/jetbrains/kotlinconf/backend/Main.kt
```diff
@@ -16,9 +16,13 @@ import io.ktor.websocket.*
...
 fun Application.main() {
-    val config = environment.config.config("service")
-    val mode = config.property("environment").getString()
+    val config = environment.config
+    val serviceConfig = config.config("service")
+    val mode = serviceConfig.property("environment").getString()
     log.info("Environment: $mode")
+    val sessionizeConfig = config.config("sessionize")
+    val sessionizeUrl = sessionizeConfig.property("url").getString()
+    val sessionizeInterval = sessionizeConfig.property("interval").getString().toLong()
```
jarekankowski: Nitpick - What do you think about creating extension function to directly get int?
```kotlin
fun ApplicationConfigValue.getLong() = getString().toLong()
```

or maybe even going a step further, we could create extensions to remove the `property("").getString()` calls?
```kotlin
fun ApplicationConfig.getPropertyString(path: String) = property(path).getString()
```

MarcinMoskala: Extension function just to replace one usage that contains single expression with two function calls... it does not sound like a good idea for me. Why designing your own API if the current one is nearly as good.

backend/src/org/jetbrains/kotlinconf/backend/Main.kt
```diff
@@ -33,6 +37,18 @@ fun Application.main() {
     install(WebSockets)
     install(XForwardedHeadersSupport)
     install(StatusPages) {
+        exception<ServiceUnavailable> { _ ->
```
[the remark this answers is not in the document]
MarcinMoskala: Thanks. I applied it previously in Kot. Academy project and this mechanism work really well. Good thing is that now you can throw this errors in any place (like in use cases or in a database) and it will be caught as well.

PR #23, opened 2018-09-10
Migrate to ktor and kotlinx.serialization

e5l:


android/app/src/main/java/org/jetbrains/kotlinconf/ui/SessionDetailsFragment.kt
```diff
@@ -38,85 +53,131 @@ class SessionDetailsFragment : Fragment(), SessionDetailsView {
...
         super.onCreateOptionsMenu(menu, inflater)
         menu.clear()
     }
 
-    override fun setRatingClickable(clickable: Boolean) {
-        goodButton.isClickable = clickable
-        okButton.isClickable = clickable
-        badButton.isClickable = clickable
+    private fun setupRatingButtons(rating: SessionRating?) {
+        goodButton.backgroundResource = if (rating == SessionRating.GOOD)
```
jarekankowski: If we extract the `if-else`:
```kotlin
if (condition)
    R.drawable.round_toggle_button_background_selected
else
    R.drawable.round_toggle_button_background
```
to a separate method, we could remove the duplication in setting the background. But considering the size of the PR and how important it is to be merged, it's definitely a nitpick. I'd recommend merging the PR first and do some cleaning later on.

android/app/src/main/res/xml/backup_descriptor.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<full-backup-content>
+    <!-- Exclude specific shared preferences that contain GCM registration Id -->
```
jarekankowski: Should we exclude favourites and votes preferences files from being backed up?

backend/build.gradle
```diff
@@ -1,77 +1,65 @@
...
 
 shadowJar {
     baseName = 'backend'
     classifier = null
     version = null
-}
\ No newline at end of file
+}
+
+kotlin.experimental.coroutines 'enable'
```
jarekankowski: You know this much better, but is it still valid as the coroutines are not experimental any more?

settings.gradle
```diff
@@ -1,10 +1,13 @@
...
-//include 'common-js'
+
+enableFeaturePreview('GRADLE_METADATA')
 
 include 'backend'
 include 'common'
-include 'common-jvm'
+include 'konfios'
+
+//include 'web' TODO: Fixme
```
jarekankowski: if I'm not mistaken, the `common-js` and `web` modules has been removed. Should we also remove the commented out `include`s?

e5l: Sure

android/app/build.gradle
```diff
@@ -43,46 +47,50 @@ android {
...
+    implementation 'net.opacapp:multiline-collapsingtoolbar:1.5.0'
+    implementation 'com.github.bumptech.glide:glide:3.7.0'
+    implementation 'com.brandongogetap:stickyheaders:0.4.9'
+
+
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:$coroutines_version"
+    implementation "io.ktor:ktor-client-android:$ktor_version"
+    implementation "io.ktor:ktor-client-gson:$ktor_version"
+
+    api "io.ktor:ktor-client-json-jvm:$ktor_version"
```
sandwwraith: You have both `io.ktor:ktor-client-gson` and `io.ktor:ktor-client-json-jvm`, is this intended?

PR #26, opened 2018-09-16
Code verification

jarekankowski:
Following our discussion with @catalinachioveanu, instead of going through the merge hell, I basically copy-pasted her implementation of code validation dialog on the top of develop. With this PR merged, we can close https://github.com/JetBrains/kotlinconf-app/pull/22.

Add a dialog to verify the user access code to vote
- the dialog appears single time when the app opens
- on the session details screen (`SessionDetailsFragment.kt`), if the code is not verified, the button with a prompt is shown
- the `submit` button is enabled only if the code is present and terms and conditions checkbox is ticked
- when app is communicating with the backend to verify the code, a progress bar is visible

PR also includes updated icons on voting buttons on the session details screen.

#### Code verification dialog
| Code only        | Checkbox only           | Code and checkbox | Loading |
| :-------------: |:-------------:| :-------------:| :-------------:|
| screenshot_1537106834 | screenshot_1537106823 | screenshot_1537106830 | screenshot_1537130099 |

#### Session details screen
| Not verified        | Verified           |
| :-------------: |:-------------:|
| screenshot_1537134264 | screenshot_1537106764 |

#### Session list screen
The time was overlapping the title. It was a single line change (https://github.com/JetBrains/kotlinconf-app/pull/26/commits/9ba236185d641057cbb0f6a1cb0ceceb3103283b) so I decided to include it to this PR.

| Before        | After           |
| :-------------: |:-------------:|
| screenshot_1537134877 | screenshot_1537134627 |

#### Gotchas
- using `themedButton` and borderless colored style for `verify voting code button` didn't work for me, I guess I was doing something wrong. To make it looking a bit nicer I set the background and the text color manually (change has been applied in https://github.com/JetBrains/kotlinconf-app/pull/26/commits/7f689b9bfe74a475bfbe06c0dbbdb4c0165de413)
- dialog jumps when keyboard is shown (small things but a bit annoying)
- what do you think if we also add a menu item to verify the code if it's not verified? Something like a small reminder to do it and it may look a bit hidden when it's only on the details screen (that's just my impression, I'm not an UX expert at all)


android/app/src/main/java/org/jetbrains/kotlinconf/model/KotlinConfViewModel.kt
```diff
@@ -1,17 +1,25 @@
...
-import android.widget.*
-import com.google.gson.*
-import io.ktor.client.call.*
-import kotlinx.coroutines.*
-import org.jetbrains.anko.*
-import org.jetbrains.kotlinconf.*
-import org.jetbrains.kotlinconf.api.*
-import org.jetbrains.kotlinconf.data.*
-import java.io.*
+import android.arch.lifecycle.LiveData
```
e5l: could you switch to star imports?

jarekankowski: yes, Sir!

MarcinMoskala:
The design looks good. About the code, I wish this branch to be merged first:
https://github.com/JetBrains/kotlinconf-app/pull/27
Then I will adjust logic on this branch to the MVP architecture (I will move logic to presenters). When I will be doing this, I will review everything else.

android/app/src/main/java/org/jetbrains/kotlinconf/model/KotlinConfViewModel.kt
```diff
@@ -48,7 +60,7 @@ class KotlinConfViewModel(
 
     init {
         loadLocalData()
-        launch {
+        launch(UI) {
```
e5l: Why we launch the update in UI thread instead of CommonPool?

PR #27, opened 2018-09-16
Move back to mpp mvp

MarcinMoskala:
Huge clean-up that introduces clean MVP again, but now both used in Android and in iOS. The repository is now used only by presenters and so there is also no need for native-specific functions. I also made generic error handling and offline mode is fully supported on both Android and iOS.


android/app/src/main/java/org/jetbrains/kotlinconf/KotlinConfApplication.kt
```diff
@@ -1,16 +1,21 @@
...
 
-import android.app.*
-import android.content.*
-import android.support.multidex.*
-import kotlinx.coroutines.*
-import kotlinx.coroutines.android.*
-import org.jetbrains.anko.*
-import org.jetbrains.kotlinconf.model.*
-import java.util.*
+import android.app.Application
```
e5l: Could you use `*` star imports everywhere instead?

common/src/main/kotlin/org/jetbrains/kotlinconf/model/KonfAppDataModel.kt
```diff
@@ -1,108 +0,0 @@
-package org.jetbrains.kotlinconf.model
```
e5l: This file leaves empty on disk :(

common/src/main/kotlin/org/jetbrains/kotlinconf/data/Session.kt
```diff
@@ -2,17 +2,18 @@ package org.jetbrains.kotlinconf.data
...
+        val isServiceSession: Boolean,
+        val isPlenumSession: Boolean,
+        val questionAnswers: List<QuestionAnswer>,
+        val speakers: List<String>,
+        val description: String?,
+        val startsAt: String,
+        val title: String,
+        val endsAt: String,
+        val categoryItems: List<Int>,
+        val roomId: Int
```
e5l: Please consider using Kotlin style guide. It could be configured in IDEA

common/src/main/kotlin/org/jetbrains/kotlinconf/presentation/CodeVerificationPresenter.kt
```diff
@@ -0,0 +1,29 @@
...
+
+class CodeVerificationPresenter(
+        private val uiContext: CoroutineContext,
+        private val view: CodeVerificationView,
+        private val repository: DataRepository
+) {
+
+    fun onCreate() {
+        launchAndCatch(uiContext, view::showError) {
+            val api = KotlinConfApi()
```
e5l: Could we avoid creating the blank API?

common/src/main/kotlin/org/jetbrains/kotlinconf/presentation/CodeVerificationPresenter.kt
```diff
@@ -0,0 +1,29 @@
...
+            api.createUser("AAAA")
```
e5l: Looks like the user-id is invalid

common/src/main/kotlin/org/jetbrains/kotlinconf/presentation/CodeVerificationPresenter.kt
```diff
@@ -0,0 +1,29 @@
...
+import kotlin.coroutines.CoroutineContext
+
+class CodeVerificationPresenter(
+        private val uiContext: CoroutineContext,
+        private val view: CodeVerificationView,
+        private val repository: DataRepository
+) {
+
+    fun onCreate() {
+        launchAndCatch(uiContext, view::showError) {
```
e5l: We should try to avoid launching a network task in the UI context(if it's possible).

common/src/main/kotlin/org/jetbrains/kotlinconf/DateFormat.kt
```diff
@@ -18,36 +18,40 @@ fun Pair<GMTDate, GMTDate>.toReadableString(): String = buildString {
...
-
 /**
  * According to mask: "yyyy-MM-dd'T'HH:mm:ss"
  */
-fun parseDate(date: String): GMTDate = date.run {
-    val year = substring(0, 4).toInt()
-    val month = substring(5, 7).toInt()
-    val day = substring(8, 10).toInt()
+fun String.parseDate(): GMTDate {
+    fun formatError(): Nothing = throw Error("Format of $this is not correct")
```
e5l: We could write a single catch for all `toInt()` and wrap exception instead.

common/src/main/kotlin/org/jetbrains/kotlinconf/api/KotlinConfApi.kt
```diff
@@ -11,7 +11,8 @@ import org.jetbrains.kotlinconf.data.*
 
 internal expect val END_POINT: String
 
-class KotlinConfApi(private val userId: String) {
```
e5l: Please consider keeping single instance creation and user state info here.
It's quite complex to manage requests and state with multiple instances.

common/src/main/kotlin/org/jetbrains/kotlinconf/data/Session.kt
```diff
@@ -2,17 +2,18 @@ package org.jetbrains.kotlinconf.data
...
-    val title: String,
-    val endsAt: String,
-    val categoryItems: List<Int>,
-    val roomId: Int
+        val id: String,
+        val isServiceSession: Boolean,
+        val isPlenumSession: Boolean,
+        val questionAnswers: List<QuestionAnswer>,
+        val speakers: List<String>,
+        val description: String?,
```
e5l: The description field is not supported on the iOS platform. Could you revert that change?

common/src/main/kotlin/org/jetbrains/kotlinconf/presentation/DataRepository.kt
```diff
@@ -4,39 +4,17 @@ import org.jetbrains.kotlinconf.*
...
-
-    /**
-     * Code verification
-     */
     suspend fun verifyCode(code: VotingCode)
-
-    /**
-     * Native callback API
-     */
-    fun update(callback: NativeCallback<Unit>)
```
e5l: The kotlin-native doesn't expose suspend functions to the iOS stubs.
So we should add alternative with callbacks for the iOS

common/src/main/kotlin/org/jetbrains/kotlinconf/presentation/SessionDetailsPresenter.kt
```diff
@@ -21,15 +21,15 @@ class SessionDetailsPresenter(
 
     fun onCreate() {
         refreshDataFromRepo()
-//        repository.addUpdateListener(onRefreshListener)
+        repository.onRefreshListeners += onRefreshListener
```
e5l: Please consider avoiding comments with code

android/app/src/main/java/org/jetbrains/kotlinconf/KotlinConfApplication.kt
MarcinMoskala: Sure

common/src/main/kotlin/org/jetbrains/kotlinconf/data/Session.kt
MarcinMoskala: Cannot, just how I commented above - this field is used to read data from Sessionize API so we cannot change it unless we decouple this two data models. Though it works fine.

common/src/main/kotlin/org/jetbrains/kotlinconf/presentation/CodeVerificationPresenter.kt
MarcinMoskala: Yes, I used it only for testing purposes.

common/src/main/kotlin/org/jetbrains/kotlinconf/DateFormat.kt
MarcinMoskala: Done

common/src/main/kotlin/org/jetbrains/kotlinconf/api/KotlinConfApi.kt
MarcinMoskala: Not sure what do you mean? Right now user can use app without user id (which is code) and after some time provide it. With this logic, we cannot keep the way it worked.

common/src/main/kotlin/org/jetbrains/kotlinconf/data/Session.kt
MarcinMoskala: As I wrote in the comment - this instance cannot be changed because it is used to parse Sessionize API. We would need to decouple those two. Though, looks like it works fine in iOS.

common/src/main/kotlin/org/jetbrains/kotlinconf/presentation/DataRepository.kt
MarcinMoskala: No, we don't. Now I use Presenters on iOS and I don't need to use repository from views (as it should be in MVP). Check out iOS. It is fully operational.

common/src/main/kotlin/org/jetbrains/kotlinconf/presentation/SessionDetailsPresenter.kt
MarcinMoskala: I always avoid them.

common/src/main/kotlin/org/jetbrains/kotlinconf/CoroutineUtils.kt
```diff
@@ -1,18 +1,30 @@
...
-        val result = try {
-            block()
-        } catch (cause: Throwable) {
-            callback(null, cause)
-            null
+fun launchAndCatch(
+        context: CoroutineContext,
+        onError: (Throwable) -> Unit,
+        function: suspend () -> Unit
+): Finallizable {
```
e5l: Finalizable

common/src/main/kotlin/org/jetbrains/kotlinconf/api/KotlinConfApi.kt
e5l: `KotlinConfApi` is stateless. Could we convert it to an object?

common/src/main/kotlin/org/jetbrains/kotlinconf/model/KonfAppDataModel.kt
MarcinMoskala: Deleted now

common/src/main/kotlin/org/jetbrains/kotlinconf/presentation/SessionDetailsPresenter.kt
MarcinMoskala: I removed comment here. I don't know who introduced it. Can we resolve this?

common/src/main/kotlin/org/jetbrains/kotlinconf/api/KotlinConfApi.kt
MarcinMoskala: We can, but it would give us nothing. It is used only in `KotlinConfDataRepository` which also have only a single instance anyway. The only change from making `KotlinConfApi` to object declaration is that it would be additionally held in a static field what are a bit challenging for Garbage Collector. Also, I am not sure how about Kotlin/Native and object declarations.

PR #34, opened 2018-09-19
Voting dialog ui

jamescoggan:
### Description
Created the UI for the Voting code.

### Todo
* We need to check if the code is already saved or not, otherwise ask with the dialog and save it.

### Screenshots
simulator screen shot - iphone 7 plus - 2018-09-19 at 00 53 01


konfios/Podfile
```diff
@@ -7,6 +7,7 @@ target 'konfswift' do
   pod 'SDWebImage', '~> 4.1.0'
   pod 'MBProgressHUD', '~> 1.0.0'
   pod 'TagListView-ObjC', '~> 0.1.1'
+  pod 'PopupDialog', '~> 0.8'
```
e5l: Should we mention new pods on the license page?

MarcinMoskala:
I think it is better if I finish on this branch instead of merging into `develop` functionality without logic.

PR #32, opened 2018-09-18
UI and functional changes to the voting dialog

jarekankowski:
##### Main Activity
- Privacy Policy acceptance is required to use the app. The `PrivacyPolicyAcceptanceFragment` is shown every time when the app is launched and the user has not accepted the Privacy Policy. The dialog is not-cancellable, which means it cannot be dismissed neither by clicking greyed-out area outside the dialog nor by pressing the back button on the keyboard.
- UI of the PP dialog has different order - the PP info has been moved to the top of the view, voting code to the bottom
- Ticked PP checkbox is required to enable the `submit` button. The voting code is optional
- Code to show the PP dialog has been moved to `onCreate()` of `MainActivity`. Calling it from the `onActivityCreated()` function of `SessionListFragment` was causing the dialog appearing twice (`AllSessionsFragment` and `FavoriteSessionsFragment` extends `SessionListFragment` and both are visible on the screen)
- Fixed link to open the PP (was linking to the T&C before)

| PP dialog | PP accepted | Code entered<br>PP not accepted | Code entered<br>PP accepted |
| :---: | :---: | :---: | :---: |
| screenshot_1537296864 | screenshot_1537296867 | screenshot_1537296889 | screenshot_1537296891 |

##### SessionDetailsFragment
- If the voting code is not verified, `VotingCodeEnterFragment` is shown when pressed a `verify code` button on the `SessionDetailsFragment`. It shows only the prompt to insert the code (no PP info)
- the voting dialog is cancellable

| Enter voting code dialog | Code entered |
| :---: | :---: |
| screenshot_1537299079 | screenshot_1537300608 |

##### InfoFragment
- `Reason to create the app` has been removed
- Link to the github project has been added after the legal notice section

| Before | After |
| :---: | :---: |
| screenshot_1537300749 | screenshot_1537297706 |
| screenshot_1537300753 | screenshot_1537298109 |

##### Minor UI changes (I'm sure everyone will notice ;)
- color primary change `#35304E` -> `#031137`
- color accent change `#F37021` -> `#FF5B00`
- the width of sessions time text view has been reduced (`SessionListFragment`) due to the `hh:mm` format (with the current implementation setting the width of the view based on the margin so please keep in mind, that if we decide to use `am/pm` the layout will look broken!)
- `voting` renamed to `rating` across the app
- added external contributors view
screenshot_1537401725

PR #35, opened 2018-09-19
Require terms acceptance

MarcinMoskala:
New logic that requires terms acceptance to use the application at all


jarekankowski:
@MarcinMoskala I've already address the change of the logic in https://github.com/JetBrains/kotlinconf-app/pull/32

PR #37, opened 2018-09-19
Require terms acceptance to use app at all

MarcinMoskala:
New logic: Now terms and conditions must be accepted to allow application usage at all. So when we open application and they are not accepted, the prompt is shown. A user cannot cancel it without accepting terms and conditions.

### DIFF
--- a/android/app/src/main/res/values/strings.xml
+++ b/android/app/src/main/res/values/strings.xml
@@ -190,4 +190,5 @@ liability.</p> ]]></string>
     <string name="code_incorrect">Sorry, the code entered is incorrect</string>
     <string name="voting_text">In order to enable voting, please enter your voting code</string>
     <string name="verify_button_text">Verify voting code</string>
+    <string name="msg_need_to_accept_terms">To use application, you need to accept the terms and conditions</string>
 </resources>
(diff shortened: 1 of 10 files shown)


android/app/src/main/res/values/strings.xml
```diff
@@ -190,4 +190,5 @@ liability.</p> ]]></string>
     <string name="code_incorrect">Sorry, the code entered is incorrect</string>
     <string name="voting_text">In order to enable voting, please enter your voting code</string>
```
hhariri: rating

android/app/src/main/res/values/strings.xml
```diff
@@ -190,4 +190,5 @@ liability.</p> ]]></string>
...
     <string name="verify_button_text">Verify voting code</string>
```
hhariri: Let's drop "voting" from there. The code is used for multiple purposes.

android/app/src/main/res/values/strings.xml
```diff
@@ -190,4 +190,5 @@ liability.</p> ]]></string>
...
+    <string name="msg_need_to_accept_terms">To use application, you need to accept the terms and conditions</string>
 </resources>
```
hhariri: To use the application, you need to accept the Privacy Policy.

jarekankowski:
@MarcinMoskala if I'm not mistaken, I've already addressed these changes in https://github.com/JetBrains/kotlinconf-app/pull/32

PR #41, opened 2018-09-21
Fix ios map

e5l:


konfios/konfswift/ui/AboutViewController.swift
```diff
@@ -51,6 +51,22 @@ class AboutViewController : UIViewController, MKMapViewDelegate {
     @IBAction func onWebsiteClick(_: Any?) {
         UIApplication.shared.openURL(URL(string: "https://www.kotlinconf.com/")!)
     }
+    
+    @IBAction func onMapClick(_: Any?) {
+//        val place = map.annotations.firstObject.uncheckedCast<MKPointAnnotation>()
```
jarekankowski: Does the commented code give any value or we can just remove it?

PR #42, opened 2018-09-23
Fix favourites after login

MarcinMoskala:


jarekankowski:
The fab still behaves quite unpredictable

ezgif com-video-to-gif 1

ezgif com-video-to-gif

MarcinMoskala:
Looks like it might be known Support Library error for v28 https://github.com/leinardi/FloatingActionButtonSpeedDial/issues/76

MarcinMoskala:
Tried their workaround and didn't work. Also tried to `invalidate` and `requestLayout` and still nothing. Though looks like moving back to Support Library version 27.1.1 fixes problem. I think we need to stay on this version until Google finally fixes it.

PR #44, opened 2018-09-24
Add prompt when the user it trying to vote

catalinachioveanu:


android/app/src/main/java/org/jetbrains/kotlinconf/ui/SessionDetailsFragment.kt
```diff
@@ -99,23 +99,20 @@ class SessionDetailsFragment : BaseFragment(), SessionDetailsView {
...
+        val favoriteIcon = if (isFavorite) R.drawable.ic_favorite_white_24dp else R.drawable.ic_favorite_border_white_24dp
+        favoriteButton.setImageResource(favoriteIcon)
+        
         session.speakers
-            .takeIf { it.size < 3 }
-            ?.map { it.profilePicture }
-            ?.apply {
-                forEachIndexed { index, imageUrl ->
-                    imageUrl?.let { speakerImageViews[index].showSpeakerImage(it) }
+                .takeIf { it.size < 3 }
```
MarcinMoskala: @e5l Asked me to change such indent from 8 to 4. I understand that this is the standard in Kotlin projects. Can you change it in your IDE and format it again?

MarcinMoskala: I won't write the same comment in other places but this applies to many lines.

PR #46, opened 2018-09-25
Remove tokens

e5l:


android/app/src/main/java/org/jetbrains/kotlinconf/ui/PrivacyPolicyAcceptanceFragment.kt
```diff
@@ -94,26 +87,6 @@ class PrivacyPolicyAcceptanceFragment : BaseDialogFragment(), CodeVerificationVi
                     view {
```
jarekankowski: It's more a design decision, but I'd consider removing the divider as well

android/app/src/main/java/org/jetbrains/kotlinconf/ui/PrivacyPolicyAcceptanceFragment.kt
```diff
@@ -18,14 +18,13 @@ import org.jetbrains.kotlinconf.*
 import org.jetbrains.kotlinconf.presentation.*
 
 
-class PrivacyPolicyAcceptanceFragment : BaseDialogFragment(), CodeVerificationView {
+class PrivacyPolicyAcceptanceFragment : BaseDialogFragment() {
     private lateinit var submitButton: Button
     private lateinit var checkBox: CheckBox
     private lateinit var codeEditText: EditText
```
jarekankowski: Looks like the `codeEditText` is not used now and can be removed

e5l:
Hi, @jarekankowski. Thanks for the review :). 
I fixed the mentioned problems, could you take a look?

PR #54, opened 2018-10-19
build: configure buildscript in root build file only

3flex:
The `buildscript {}` block should be used in the root build file only (see [this recent advice](https://discuss.gradle.org/t/multiproject-gradle-seems-to-need-plugin-repository-declared-in-each-subproject/27970/4) from Gradle core dev).

This PR removes all applications of `buildscript {}` from subprojects as recommended.


e5l:
Hi, @3flex. Thanks for the PR. LGTM

PR #88, opened 2024-01-16
Load dark resources in night mode

e5l:


terrakok:
what is a difference between ligh/dark xml icons? If it is only a tone than it can be tinted in the runtime.

terrakok:
Usually dark/light resources are needed for non-vectored icons

e5l:
Hey @terrakok, thank you for the review!

Sure, the icons are the same and tinted at runtime. I have an idea how this can be fixed!

terrakok (approved):

PR #91, opened 2024-01-30
Migrate to the new resources API

e5l:


shared/src/mobileMain/kotlin/org/jetbrains/kotlinconf/ui/components/AsyncImage.kt
```diff
@@ -14,22 +14,22 @@ import androidx.compose.ui.Modifier
...
 
-@OptIn(ExperimentalResourceApi::class)
 @Composable
 fun AsyncImage(
     modifier: Modifier,
     imageUrl: String,
     contentDescription: String,
     contentScale: ContentScale = ContentScale.Crop,
 ) {
+    remember {  }
```
terrakok, line 28: outdated

shared/src/mobileMain/kotlin/org/jetbrains/kotlinconf/ui/components/Tag.kt
```diff
@@ -46,17 +56,20 @@ fun Tag(
     }
 }
 
+@OptIn(ExperimentalResourceApi::class)
 @Composable
 fun LightningTalk(title: String, dimmed: Boolean = false) {
-    Tag(Icons.LIGHT, title, dimmed)
+    Tag(Res.drawable.light.painter(), title, dimmed)
```
terrakok, line 62: here and in other places I see you use a painter as a function parameter.
I would say that to pass a drawable is better and get painter in the function inside

shared/src/mobileMain/kotlin/org/jetbrains/kotlinconf/ui/AboutConf.kt
```diff
@@ -245,7 +247,7 @@ private fun Labs() {
 
         Row(Modifier.padding(16.dp)) {
             Icon(
-                painter = Icons.AWS_LAB,
+                painter = painterResource(Res.drawable.aws_lab),
```
terrakok, line 250: to be equal with other code I'd expect the `.painter()` here

terrakok (changes requested):

shared/src/mobileMain/kotlin/org/jetbrains/kotlinconf/ui/components/AsyncImage.kt
e5l: thank you, fixed

shared/src/mobileMain/kotlin/org/jetbrains/kotlinconf/ui/components/Tag.kt
e5l: refactored

gradle.properties
```diff
@@ -15,3 +15,4 @@ development=true
 kotlin.mpp.stability.nowarn=true
 kotlin.mpp.enableCInteropCommonization=true
 kotlin.mpp.androidSourceSetLayoutVersion=2
+compose.resources.always.generate.accessors=true
```
terrakok, line 18: this is not required here. the plugin will generate Res class automatically

e5l: will drop, thank you

PR #92, opened 2024-01-31
Add svg map with gestures (with fake map for now)

e5l:


shared/src/mobileMain/kotlin/org/jetbrains/kotlinconf/ui/LocationScreen.kt
```diff
@@ -0,0 +1,35 @@
...
+            tabs = listOf("FLOOR -1", "FLOOR 0", "FLOOR 1"),
+            selected = floor,
+            onSelect = { floor = it },
+        )
+
+    }
+}
+
+@Composable
+expect fun SvgView(modifier: Modifier)
```
terrakok, line 35: I would name it `SvgMapView`(filePath: String, modifier: Modifier)

terrakok (approved):

shared/src/mobileMain/kotlin/org/jetbrains/kotlinconf/ui/LocationScreen.kt
e5l: thanks, will fix

PR #93, opened 2024-02-01
Update Privacy Policy Screen

e5l:


terrakok (changes requested):

e5l:
I will move them to the resource as soon we have a new version for this year. Will also change source format to markdown

e5l:
Will be moved in a separate PR as agreed

PR #104, opened 2024-02-29
Fix tab bar button is not active when selected

e5l:


shared/src/mobileMain/kotlin/org/jetbrains/kotlinconf/App.kt
```diff
@@ -13,7 +13,8 @@ fun App(context: ApplicationContext) {
     KotlinConfTheme {
         val service = ConferenceService(
             context,
-            "https://kotlin-conf-staging.labs.jb.gg/"
+            "http://10.0.2.2:8080/"
```
terrakok, line 16: debug?

terrakok (approved):

shared/src/mobileMain/kotlin/org/jetbrains/kotlinconf/App.kt
e5l: Will use it until the backend is not deployed

PR #121, opened 2024-05-10
Add desktop support

e5l:


shared/src/jvmMain/kotlin/org/jetbrains/kotlinconf/utils/MobileLogger.jvm.kt
```diff
@@ -0,0 +1,9 @@
+package org.jetbrains.kotlinconf.utils
+
+import io.ktor.client.plugins.logging.Logger
+
+actual fun mobileLogger(): Logger = object : Logger {
```
terrakok, line 5: appLogger?

shared/src/jvmMain/kotlin/org/jetbrains/kotlinconf/utils/Dispatchers.jvm.kt
```diff
@@ -0,0 +1,7 @@
+package org.jetbrains.kotlinconf.utils
+
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Dispatchers
+
+actual val Dispatchers.App: CoroutineDispatcher
```
terrakok, line 6: As I understand we may use Dispatchers.Main here

shared/src/jvmMain/kotlin/org/jetbrains/kotlinconf/storage/ApplicationStorage.jvm.kt
```diff
@@ -0,0 +1,59 @@
...
+            }
+
+            GlobalScope.launch(Dispatchers.IO) {
+                data.debounce(5_000)
+                    .collect { storageData ->
+                    val content = Json.encodeToString(storageData)
+                    runCatching {
+                        File("store.json").writeText(content)
+                    }
+                }
```
terrakok, line 40: I'd say 5 sec is too much. And code formatting is broken here.
Could you extract file to the property and use standard app directory on user's home?

terrakok (changes requested):

shared/src/jvmMain/kotlin/org/jetbrains/kotlinconf/storage/ApplicationStorage.jvm.kt
e5l: I'm unsure if using the user directory is a good idea, cause we provide no way to uninstall it.

shared/src/jvmMain/kotlin/org/jetbrains/kotlinconf/utils/MobileLogger.jvm.kt
e5l: thanks, fixed

shared/src/jvmMain/kotlin/org/jetbrains/kotlinconf/utils/Dispatchers.jvm.kt
e5l: The `Dispatchers.Main` purpose is UI updates. This dispatcher is used for IO operations

terrakok (approved):

PR #122, opened 2024-05-12
Introduce Web Version

bashor:


shared/build.gradle.kts
```diff
@@ -11,6 +15,28 @@ plugins {
 kotlin {
     androidTarget()
     jvm()
+    
+    wasmJs {
+        binaries.executable()
+        browser { 
+
```
e5l, line 22: empty line (block)

shared/build.gradle.kts
```diff
@@ -11,6 +15,28 @@ plugins {
...
+        }
+    }
+
+    js {
+        binaries.executable()
+        browser { 
+            commonWebpackConfig {
+                val out = this.output
```
e5l, line 30: Please log an issue and add link in comment

e5l (changes requested):
Please exclude all personal photos from the repository

bashor:
> Please exclude all personal photos from the repository
Why?

e5l:
We don't have consent

bashor:
Aren't they public? Otherwise, how can we even use them in the app?

shared/build.gradle.kts
bashor: It's not a bug, it was required for dynamic loading.

Removed, since right now it's not needed, likely will return later.

shared/build.gradle.kts
bashor: empty line is removed

e5l:
Awesome!

e5l:
Reverted due to backend fail

PR #128, opened 2024-05-17
Add notification manager for web

JSMonk:


shared/src/mobileMain/kotlin/org/jetbrains/kotlinconf/ui/SearchScreen.kt
```diff
@@ -82,11 +82,11 @@ fun SearchScreen(
                     selectedTab = it
                 }
             )
-            Row(horizontalArrangement = Arrangement.End, modifier = Modifier.fillMaxWidth()) {
+            Row(horizontalArrangement = Arrangement.Start, modifier = Modifier.fillMaxWidth()) {
```
e5l, line 85: It looks like it affects the mobile design as well. Should it be platform specific?

e5l: btw. UI changes should not be related to the notification manager

e5l (changes requested):

shared/src/webMain/kotlin/org/jetbrains/kotlinconf/NotificationManager.web.kt
```diff
@@ -1,17 +1,62 @@
...
+
 actual class NotificationManager actual constructor(
     context: ApplicationContext
 ) {
+    private var notificationAllowed = false
+
     actual fun requestPermission() {
+        Notification.requestPermission {
+            if (it == "granted") {
+                notificationAllowed = true
```
bashor, line 43: Notification will be shown only on the first open, after reopen/refresh `notificationAllowed` will be `false`.

JSMonk: As far as I remember, it should be saved by browser for a specific host

bashor: But it's runtime value, and it will not be persisted between runs

shared/src/webMain/kotlin/org/jetbrains/kotlinconf/NotificationManager.web.kt
```diff
@@ -1,17 +1,62 @@
...
+            title: title,
+          });
+        });
+        }"""
+    )
+
 actual class NotificationManager actual constructor(
     context: ApplicationContext
 ) {
+    private var notificationAllowed = false
```
bashor, line 38: It should be initialized from `Notification.permission`

bashor:
Pushed Notification related things to the master without UI changes

PR #129, opened 2024-05-17
Optimize wasm files with binaryen when building the final web app bundle

bashor:


shared/build.gradle.kts
```diff
@@ -197,6 +197,11 @@ val buildWebApp by tasks.creating(Copy::class) {
     val jsWebpack = "jsBrowserProductionWebpack"
 
     dependsOn(wasmWebpack, jsWebpack)
+    
+    // TODO could be removed after migration to Kotlin 2.0+
+    kotlin.wasmJs {
```
e5l, line 202: feel free to migrate using 2.0.0 latest RC

e5l (approved):

shared/build.gradle.kts
bashor: It's better to do separately, since may require updating some other things

PR #153, opened 2024-11-04
Add colors and styles, initial drafts of components

zsmb13:
Adding the colors and text styles from the design.

I've also created a couple of simple components with some common previews. These are not meant to be production-ready yet, but they showcase how to access theme values.

### DIFF
--- a/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui25/components/KCText.kt
+++ b/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui25/components/KCText.kt
@@ -1,0 +1,23 @@
+package org.jetbrains.kotlinconf.ui25.components
+
+import androidx.compose.foundation.text.BasicText
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.text.TextStyle
+import org.jetbrains.kotlinconf.ui25.theme.KotlinConfTheme
+
+@Composable
+fun KCText(
+    text: String,
+    modifier: Modifier = Modifier,
+    color: Color = KotlinConfTheme.colors.primaryText,
+    style: TextStyle = KotlinConfTheme.typography.text1,
+) {
+    BasicText(
+        text = text,
+        modifier = modifier,
+        style = style,
+        color = { color }
+    )
+}
--- a/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui25/components/NowLabel.kt
+++ b/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui25/components/NowLabel.kt
@@ -1,0 +1,42 @@
+package org.jetbrains.kotlinconf.ui25.components
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.shape.CircleShape
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.unit.dp
+import org.jetbrains.compose.ui.tooling.preview.Preview
+import org.jetbrains.kotlinconf.ui25.theme.KotlinConfTheme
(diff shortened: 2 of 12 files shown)


shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui25/components/KCText.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import androidx.compose.foundation.text.BasicText
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.text.TextStyle
+import org.jetbrains.kotlinconf.ui25.theme.KotlinConfTheme
+
+@Composable
+fun KCText(
```
e5l, line 11: Could you tell me if this is a name from the design? If not, I would try naming it by the place it will be used

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui25/components/NowLabel.kt
```diff
@@ -0,0 +1,42 @@
...
+            text = "Now",
+            color = KotlinConfTheme.colors.accentText,
+            style = KotlinConfTheme.typography.text2
+        )
+    }
+}
+
+@Composable
+@Preview
+fun NowLabelPreview() {
```
e5l, line 38: please consider making previews private, so they will not display in completion

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui25/components/SwitcherItem.kt
```diff
@@ -0,0 +1,80 @@
...
+import org.jetbrains.kotlinconf.ui25.theme.KotlinConfTheme
+import org.jetbrains.kotlinconf.ui25.theme.PreviewHelper
+
+
+enum class SwitcherItemState {
+    Active, Normal, Selected,
+}
+
+@Composable
+fun SwitcherItem(
```
e5l, line 24: Could you be also more specific with naming? If we have a second switcher, we could struggle to distinguish between them

e5l (approved):
Nicely done! lgtm

Please check some comments before merging :)

kropp (approved):

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui25/components/KCText.kt
zsmb13: This doesn't exist as a component in the design itself, it's only here to ease the implementation. Since we don't want to use Material `Text` directly, we need to drop down to `BasicText` instead, which for some reason takes a `ColorProducer` function as a parameter instead of a simple `Color` value.

So this component helps us create text and pass in colour to it easily, while also defaulting to the most common look of `KotlinConfTheme`. We can certainly add some documentation on the component.

Not sure what to name it though. Ideally, we'd use the `Text` name for it (as our design system substitues Material), but that only really works once we've removed the dependency on Material, otherwise it'll be a headache to import the correct one.

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui25/components/SwitcherItem.kt
zsmb13: This is the name of the component in the design.

image

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui25/components/NowLabel.kt
zsmb13: Didn't realize they can be private. Fixed!

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui25/components/SwitcherItem.kt
e5l: lgtm

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui25/components/KCText.kt
e5l: We can try something like `BaseText` or `StyledText`

zsmb13: Renamed to `StyledText` and added a comment to explain it

zsmb13:
This fixed #149

PR #174, opened 2024-11-18
Add hot reload plugin

zsmb13:


ui-components/src/jvmMain/kotlin/org/jetbrains/kotlinconf/ui/components/Gallery.kt
```diff
@@ -0,0 +1,48 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.unit.Density
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.window.Window
+import androidx.compose.ui.window.application
+import androidx.compose.ui.window.rememberWindowState
+import org.jetbrains.compose.reload.DevelopmentEntryPoint
+
+fun main() {
```
sellmair, line 17: If you do not want this code to be in your production sources, feel free to put them into the 'jvmDev' source set 👍 ! I think launching the code then requires the IDE plugin from Kostya or a cli command that I could provide!

settings.gradle.kts
```diff
@@ -9,6 +9,7 @@ pluginManagement {
         google()
         gradlePluginPortal()
         mavenCentral()
+        maven("https://repo.sellmair.io")
```
sellmair, line 12: I should try to push the next builds to JetBrains infrastructure 👀

sellmair (approved):

ui-components/src/jvmMain/kotlin/org/jetbrains/kotlinconf/ui/components/Gallery.kt
e5l: we also could have a Gradle task for this in the plugin :)

e5l (approved):
Looks nice! LGTM

e5l:
Please consider adding the PR description

ui-components/src/jvmMain/kotlin/org/jetbrains/kotlinconf/ui/components/Gallery.kt
sellmair: Oh there is a task: 

```
gw :app:devRun -DclassName=my.pkg.ClassNameKt -DfunName=MyComposable
```

zsmb13: I definitely prefer an easily clickable `main` function for the time being, just to keep things super simple, but we can revisit this later. A saved run config would make it fairly painless to launch the task above, for example.

I'm also still thinking about what entry point to provide exactly though, whether it should be a bunch of small ones for components, or if we should have a "gallery app" entry point where you can go through all of them.

sellmair: Absolutely fair!

sellmair:
@zsmb13 : Just wanted to let you know, that I published new builds
https://github.com/JetBrains/kotlinconf-app/commit/d3836a9e6d65fdf00471c9c466a194468262a85f

This will be dev-28 for hot reload and 2.0.21-firework.28 for Kotlin

PR #192, opened 2024-11-23
Add Kodee images

zsmb13:
image


e5l (approved):
Could you tell me why we need PNG versions and if we can replace them with vectors?

zsmb13:
> Could you tell me why we need PNG versions and if we can replace them with vectors?

I couldn't export these specific ones as vectors from the design. Those three images are the ones on the feedback form. I'll double-check before merging to see if we can vectorize these as well.

PR #195, opened 2024-11-23
Draw thumb color using a canvas

StylianosGakis:
This avoids any possible future problems with the .border modifier drawing a white artifact around the edges of the thumb circle

From discussion:
https://kotlinlang.slack.com/archives/CJLTWPH7S/p1732394481077219?thread_ts=1732393581.144109&cid=CJLTWPH7S

### DIFF
--- a/ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/Toggle.kt
+++ b/ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/Toggle.kt
@@ -89,7 +89,7 @@ fun Toggle(
                     indication = LocalIndication.current,
                 )
                 .drawBehind {
-                  drawCircle(color = thumbCenterColor, radius = (size.minDimension / 2) - 2.dp.toPx())
+                    drawCircle(color = thumbCenterColor, radius = (size.minDimension / 2) - 2.dp.toPx())
                 }
         )
     }
--- a/ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/Toggle.kt
+++ b/ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/Toggle.kt
@@ -22,6 +22,8 @@ import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
+import androidx.compose.ui.draw.drawBehind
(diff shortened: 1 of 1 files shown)


ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/Toggle.kt
```diff
@@ -72,20 +74,23 @@ fun Toggle(
...
                 .clip(ToggleThumbShape)
+                .background(toggleColor)
                 .clickable(
                     onClick = { onToggle(!enabled) },
                     interactionSource = interactionSource,
                     indication = LocalIndication.current,
                 )
-                .background(KotlinConfTheme.colors.mainBackground)
+                .drawBehind {
+                  drawCircle(color = thumbCenterColor, radius = (size.minDimension / 2) - 2.dp.toPx())
```
zsmb13, line 92:
```suggestion
                    drawCircle(color = thumbCenterColor, radius = (size.minDimension / 2) - 2.dp.toPx())
```

zsmb13 (approved):
Looks good, thank you!

ColtonIdle:
Just here to learn. 😅 
I'm confused why @StylianosGakis changes are "better" than what we already had. Didn't adding the `              .wrapContentSize(unbounded = true)` get rid of the anti aliasing?

ColtonIdle:
>  effectively rendering outside of the size it takes up, hence the unbounded = true part.
Hm. I'll need to play around with that. i thought the view system clipped, but compose (by default) did not clip when drawn out of bounds

>  I simply did a PR to avoid using the border and do the drawing using a canvas, that's all.
got it. thanks for teaching!

StylianosGakis:
It does not clip. But if you do .size(999.dp) on a composable which previously had its constraints set to something like `Constraints(minWidth = 0, maxWidth = 50, minHeight = 0, maxHeight = 50)` then you will not break out of those constraints. They will be respected.
what `wrapContentX` does is that it wraps the content so that it only takes up as much space as it really needs, and adding `unbounded = true` to it also changes the constraints downstream by turning them to `Constraints(minWidth = 0, maxWidth = Infinite, minHeight = 0, maxHeight = Infinite)`. So you can now once again try to grow bigger than what you previously could.

Take a look at the impl here https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/foundation/foundation-layout/src/commonMain/kotlin/androidx/compose/foundation/layout/Size.kt;l=1010?q=private%20class%20WrapContentNode

PR #201, opened 2024-11-25
Add MainHeader component

### LINKED ISSUE #196: Create Main Header Component

zsmb13:
Fixes #196 

Covers all different variants of the main header:

https://github.com/user-attachments/assets/d20ebb24-9bdc-4e90-905c-c9210dbec286

See the usage examples for the public API in the Preview function code.


ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/Gallery.kt
```diff
@@ -44,7 +45,6 @@ fun GalleryApp() {
                 SwitcherPreview()
                 TogglePreview()
                 TopMenuButtonPreview()
-                TopMenuTitlePreview()
```
zsmb13, line 47: Removed this component, as it was super simple, and only used as part of `MainHeader`. We can probably do this for a couple more similar components later.

ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/NowButton.kt
```diff
@@ -30,18 +39,24 @@ private val NowButtonShape = RoundedCornerShape(
     bottomEndPercent = 50,
 )
 
+enum class NowButtonState {
+    Before, Current, After,
```
zsmb13, line 43: Changed the `NowButton` from 2 possible states to 3, as we might need to scroll either up or down, depending on which way the current point in time is within the schedule.

image

e5l:
recording doesn't work

e5l (approved):

ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/MainHeader.kt
```diff
@@ -0,0 +1,258 @@
...
+                onValueChange = { onSearchValueChange(it) },
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .focusRequester(focusRequester),
+                singleLine = true,
+                textStyle = KotlinConfTheme.typography.text1
+                    .copy(color = KotlinConfTheme.colors.primaryText),
+                cursorBrush = SolidColor(KotlinConfTheme.colors.primaryText),
+            )
+            androidx.compose.animation.AnimatedVisibility(
```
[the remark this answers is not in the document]
zsmb13: Sadly, that doesn't solve the issue, it would still be ambiguous.

The problem is that there are several functions with the exact same import, `androidx.compose.animation.AnimatedVisibility`:

- [`AnimatedVisibility()`](https://developer.android.com/reference/kotlin/androidx/compose/animation/package-summary#AnimatedVisibility(kotlin.Boolean,androidx.compose.ui.Modifier,androidx.compose.animation.EnterTransition,androidx.compose.animation.ExitTransition,kotlin.String,kotlin.Function1))
- [`RowScope.AnimatedVisibility()`](https://developer.android.com/reference/kotlin/androidx/compose/animation/package-summary#(androidx.compose.foundation.layout.RowScope).AnimatedVisibility(kotlin.Boolean,androidx.compose.ui.Modifier,androidx.compose.animation.EnterTransition,androidx.compose.animation.ExitTransition,kotlin.String,kotlin.Function1))
- [`ColumnScope.AnimatedVisibility()`](https://developer.android.com/reference/kotlin/androidx/compose/animation/package-summary#(androidx.compose.foundation.layout.ColumnScope).AnimatedVisibility(kotlin.Boolean,androidx.compose.ui.Modifier,androidx.compose.animation.EnterTransition,androidx.compose.animation.ExitTransition,kotlin.String,kotlin.Function1))

So whether named or not, if here we call the function with the `AnimatedVisibility()` syntax, it will resolve to the extension on `RowScope` instead of the top level function. The fully qualified name prevents it from using the implicit receiver.

Here's a [StackOverflow question](https://stackoverflow.com/questions/67975569/why-cant-i-use-animatedvisibility-in-a-boxscope) on the same problem, and a [YouTrack issue](https://youtrack.jetbrains.com/issue/KT-48215).

PR #206, opened 2024-12-01
Add FeedbackForm component

### LINKED ISSUE #205: Create Feedback Form component

zsmb13:
Fixes #205 

https://github.com/user-attachments/assets/4ae7c8e6-bb03-4774-94ab-83ac59d4e11e

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -10,11 +10,11 @@ androidx-test-junit = "1.2.1"
 androidx-work-runtime = "2.9.1"
 androidx-preference = "1.2.1"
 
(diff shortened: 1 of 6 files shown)


e5l (approved):
Looks nice!

Could you check how it works with an onscreen keyboard on a device? It would be nice to have a video :)

zsmb13:
@e5l The challenges with on-screen keyboards will be about managing the surrounding scrolling layout and the insets of the entire app, so I think we'll have to check this later. I'm pretty sure there will be work to do there, but it will certainly come up when building actual screens that use inputs.

PR #215, opened 2024-12-19
Add stub screens for navigation

kropp:
Depends on #214

### DIFF
--- a/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/App.kt
+++ b/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/App.kt
@@ -7,10 +7,13 @@ import androidx.compose.animation.core.tween
 import androidx.compose.animation.slideInHorizontally
 import androidx.compose.animation.slideOutHorizontally
 import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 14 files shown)


shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/Model.kt
```diff
@@ -3,9 +3,22 @@ package org.jetbrains.kotlinconf
...
+@JvmInline
+value class SpeakerId(val id: String) {
+    override fun toString(): String = id
+}
+
+@Serializable
+@JvmInline
+value class SessionId(val id: String) {
+    override fun toString(): String = id
+}
```
zsmb13, line 20: These new wrapper classes broke the `backend` module, the CI check is failing to compile

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/AboutApp.kt
```diff
@@ -0,0 +1,21 @@
...
+import kotlinconfapp.shared.generated.resources.arrow_left_24
+import org.jetbrains.compose.resources.painterResource
+import org.jetbrains.kotlinconf.LocalNavController
+import org.jetbrains.kotlinconf.ui.components.StyledText
+
+@Composable
+fun AboutApp() {
+    val navController = LocalNavController.current
+    Column {
+        Image(painterResource(Res.drawable.arrow_left_24), "back", modifier = Modifier.clickable { navController.popBackStack() })
```
zsmb13, line 18: What's your take on using `popBackStack` vs `navigateUp` in general?

This video reasons for using `navigateUp` instead of `popBackStack`: https://www.youtube.com/watch?v=y2zLFONuk7c

zsmb13 (changes requested):

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/App.kt
```diff
@@ -1,29 +1,128 @@
...
+                    startDestination = StartScreen,
+                    modifier = Modifier.fillMaxSize(),
+                    enterTransition = enterTransition { it },
+                    exitTransition = exitTransition { -it },
+                    popEnterTransition = enterTransition { -it },
+                    popExitTransition = exitTransition { it },
+                ) {
+                    composable<StartScreen> {
+                        Column {
+                            StyledText("Start screen")
```
e5l, line 66: Please consider extracting string constants

e5l (changes requested):

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/Model.kt
kropp: Fixed in #214

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/AboutApp.kt
kropp: Thanks for the link, haven't thought about it before in that way.
As many commenters have pointed out, the issue is not with `popBackStack`/`navigateUp`, but rather with preventing duplicate clicks or, in general, any user actions during the navigation transition.
As our own Ivan Matkov suggested in the comments there, we may check Lifecycle state to prevent this.

On a side note, the Back button behaves differently on iOS, as there you're always navigating inside the app, while on Android with `navigateUp` you may get back to a previous app, depending on how our app was launched.

zsmb13: Ah yes, I've also heard the solution with checking the lifecycle. I think the core issue is that people want an API that's a single function and safe to use by default.

I guess they can all write their own `popBackStackSafely()` functions, which are `@Composable` so that they can reach out to the lifecycle to check it before doing a pop.

zsmb13 (approved):
Added a small bugfix, the rest looks good to me

PR #223, opened 2025-01-14
Add start screens, privacy policy and notification onboarding

zsmb13:
https://github.com/user-attachments/assets/496d0322-05bd-444f-a3ae-204af536ab59


shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/View.kt
```diff
@@ -21,10 +17,10 @@ data class Speakers(
 }
 
 @OptIn(ExperimentalResourceApi::class)
-enum class EventDay(val title: StringResource) {
-    May22(Res.string.day_1),
-    May23(Res.string.day_2),
-    May24(Res.string.day_3);
```
zsmb13, line 27: Removing these hardcoded day strings for now. We should see if these can be computed more reasonably from the backend data.

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/App.kt
```diff
@@ -36,78 +37,109 @@ import org.jetbrains.kotlinconf.screens.Speaker
...
+                        popExitTransition = exitTransition { it },
+                    ) {
+                        composable<StartScreen> {
+                            Column {
+                                StyledText("Start screen")
+                                StyledText(
+                                    "Privacy policy",
+                                    Modifier.clickable { navController.navigate(StartPrivacyPolicyScreen) }
+                                )
+                                StyledText("About App", Modifier.clickable { navController.navigate(AboutAppScreen) })
```
e5l, line 75: Please consider extracting string constant to resources

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/View.kt
e5l: Sometimes formatting of days is changed in the sessionize for the website requirements, so we need to compute this from the actual sessions timestamps

e5l (approved):

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/StartScreens.kt
```diff
@@ -0,0 +1,224 @@
...
+                        onClick = { detailsVisible = true }
+                    )
+                }
+                Spacer(Modifier.weight(1f))
+            }
+        }
+
+        Row(
+            horizontalArrangement = Arrangement.spacedBy(16.dp),
+            modifier = Modifier.padding(horizontal = 12.dp, vertical = 16.dp)
```
kropp, line 130: As I started implementing some screens as well, I noticed this `12.dp` padding everywhere.
Should we probably extract it to the theme object?

kropp (approved):

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/StartScreens.kt
zsmb13: I would only do it if the Figma design can also use it across the entire design, and show it as a named variable. Otherwise it'd be guesswork when this is "the same margin value" and when it isn't.

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/App.kt
zsmb13: All of this in `App.kt` is placeholder buttons at the moment, just to be able to open screens until we have the main nav built out, so it'll be removed anyway

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/StartScreens.kt
kropp: Fair enough

PR #226, opened 2025-01-15
add General Terms and Conditions and Privacy Policy for visitors screens

kropp:
There are some minor issues with Markdown rendering (#227), but the screens themselves are ready.

### DIFF
--- a/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/TermsOfUse.kt
+++ b/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/TermsOfUse.kt
@@ -1,21 +1,24 @@
 package org.jetbrains.kotlinconf.screens
 
-import androidx.compose.foundation.Image
-import androidx.compose.foundation.clickable
-import androidx.compose.foundation.layout.Column
 import androidx.compose.runtime.Composable
-import androidx.compose.ui.Modifier
 import kotlinconfapp.shared.generated.resources.Res
-import kotlinconfapp.shared.generated.resources.arrow_left_24
-import org.jetbrains.compose.resources.painterResource
-import org.jetbrains.kotlinconf.LocalNavController
-import org.jetbrains.kotlinconf.ui.components.StyledText
+import kotlinconfapp.shared.generated.resources.general_terms
+import kotlinconfapp.shared.generated.resources.visitors_terms_title
+import kotlinconfapp.shared.generated.resources.visitors_terms_version
+import org.jetbrains.compose.resources.ExperimentalResourceApi
+import org.jetbrains.compose.resources.stringResource
+import org.jetbrains.kotlinconf.MarkdownScreenWithTitle
 
 @Composable
-fun TermsOfUse() {
-    val navController = LocalNavController.current
-    Column {
-        Image(painterResource(Res.drawable.arrow_left_24), "back", modifier = Modifier.clickable { navController.popBackStack() })
-        StyledText("Terms of Use")
-    }
+fun TermsOfUse(onBack: () -> Unit) {
+    MarkdownScreenWithTitle(
+        title = stringResource(Res.string.general_terms),
+        header = stringResource(Res.string.visitors_terms_title),
+        subheader = stringResource(Res.string.visitors_terms_version),
+        loadText = {
+            @OptIn(ExperimentalResourceApi::class)
+            Res.readBytes("files/visitors-terms.md")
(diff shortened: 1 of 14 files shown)


ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/MarkdownScreenWithTitle.kt
```diff
@@ -0,0 +1,105 @@
...
+import com.mikepenz.markdown.model.DefaultMarkdownColors
+import com.mikepenz.markdown.model.DefaultMarkdownTypography
+import kotlinconfapp.ui_components.generated.resources.Res
+import kotlinconfapp.ui_components.generated.resources.arrow_left_24
+import kotlinconfapp.ui_components.generated.resources.main_header_back
+import org.jetbrains.compose.resources.stringResource
+import org.jetbrains.kotlinconf.ui.theme.KotlinConfTheme
+
+@Composable
+fun MarkdownScreenWithTitle(
```
zsmb13, line 29: Since this is not an individual component but really an entire screen, I'd say it's not part of the design system but the app code. It's great to have it extracted and reused like this, I'd just move it to the `shared` module instead of having it in `ui-components`.

ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/MarkdownScreenWithTitle.kt
```diff
@@ -0,0 +1,105 @@
...
+                .verticalScroll(rememberScrollState())
+        ) {
+
+            StyledText(header, style = KotlinConfTheme.typography.h1, modifier = Modifier.padding(12.dp))
+            if (subheader.isNotEmpty()) {
+                StyledText(
+                    subheader,
+                    style = KotlinConfTheme.typography.text2,
+                    color = KotlinConfTheme.colors.noteText,
+                    modifier = Modifier.padding(horizontal = 12.dp)
```
zsmb13, line 63: You could lift these `12.dp` horizontal paddings to the containing `Column` to simplify a tiny bit

ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/MarkdownScreenWithTitle.kt
```diff
@@ -0,0 +1,105 @@
...
+                )
+            }
+
+            var text by remember { mutableStateOf("") }
+            LaunchedEffect(loadText) {
+                text = loadText().decodeToString()
+            }
+
+            Markdown(
```
zsmb13, line 72: We need Markdown rendering for other screens as well (like the copy of the privacy policy that's a special startup screen). Could you extract this nicely set up `Markdown` call into a separate component that we can reuse on its own, by just passing in the text?

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/TermsOfUse.kt
```diff
@@ -1,21 +1,24 @@
...
-        StyledText("Terms of Use")
-    }
-}
\ No newline at end of file
+    MarkdownScreenWithTitle(
+        title = stringResource(Res.string.general_terms),
+        header = stringResource(Res.string.visitors_terms_title),
+        subheader = stringResource(Res.string.visitors_terms_version),
+        loadText = { Res.readBytes("files/visitors-terms.md") },
+        onBack = navController::popBackStack
```
zsmb13, line 24: The `NavController` passed down through composition locals makes me uncomfortable, as it tightly couples the individual screens to the nav setup, and also spreads pieces of navigation logic all over these screens.

What do you think about routing these navigation events back up to the `App` level so that they can be handled there, centrally?

It's what I did for screens on https://github.com/JetBrains/kotlinconf-app/pull/223/files

zsmb13 (changes requested):
Looks good, a couple small asks

ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/MarkdownScreenWithTitle.kt
kropp: Done, and added to that Start screen as well

zsmb13 (approved):

PR #229, opened 2025-01-15
Screens/code of conduct

kropp:
(on top of #226)


shared/src/commonMain/composeResources/values/strings.xml
```diff
@@ -12,6 +12,8 @@ We will process your data in accordance with the App Privacy Policy. You can adj
     <string name="notifications_description">Customize your preferences to ensure you receive only the notifications that matter most to you, avoiding unnecessary disruptions.</string>
     <string name="notifications_lets_get_started">Let’s get started!</string>
 
+    <string name="code_of_conduct">Code of conduct</string>
```
zsmb13, line 15:
```suggestion
    <string name="code_of_conduct">Code of Conduct</string>
```

I see this spelling in both the design and the content doc

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/PrivacyPolicyForVisitors.kt
```diff
@@ -20,5 +24,7 @@ fun PrivacyPolicyForVisitors() {
         subheader = stringResource(Res.string.privacy_policy_for_visitors_version),
         loadText = { Res.readBytes("files/visitors-privacy-policy.md") },
         onBack = navController::popBackStack
-    )
+    ) {
+        Spacer(Modifier.height(24.dp))
+    }
```
zsmb13, line 29: Maybe this should be the default value for this parameter? There could also be padding that's always added at the end of the content (as I see it, we add it in all 3 usages so far), and then the default for this parameter could be `{}`.

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/CodeOfConduct.kt
```diff
@@ -1,21 +1,37 @@
...
 fun CodeOfConduct() {
     val navController = LocalNavController.current
-    Column {
-        Image(painterResource(Res.drawable.arrow_left_24), "back", modifier = Modifier.clickable { navController.popBackStack() })
-        StyledText("Code of Conduct")
+    MarkdownScreenWithTitle(
+        title = stringResource(Res.string.code_of_conduct),
+        header = stringResource(Res.string.code_of_conduct),
+        loadText = { Res.readBytes("files/code-of-conduct.md") },
+        onBack = navController::popBackStack,
```
zsmb13, line 27: Same question about navigation as on the other PR https://github.com/JetBrains/kotlinconf-app/pull/226#discussion_r1917402887

zsmb13 (changes requested):

zsmb13:
PS. I changed the base branch of this PR to the branch it's based on. Once the previous PR is merged and that branch is deleted, the base branch here should automatically switch to the `kotlinconf-2025` branch. In the meantime, we see a cleaner diff here.

shared/src/commonMain/composeResources/values/strings.xml
kropp: The casing is inconsistent in design mockups, asked for clarification

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/PrivacyPolicyForVisitors.kt
kropp: I want to revisit bottom paddings later

zsmb13 (approved):

PR #239, opened 2025-01-18
Create licenses screen

zsmb13:
https://github.com/user-attachments/assets/8ec5424e-4642-4537-89c4-a240db4b8fc5

https://github.com/user-attachments/assets/31341c1d-b116-4748-98ed-264dee3aac5e


build.gradle.kts
```diff
@@ -15,4 +16,4 @@ tasks.wrapper {
     gradleVersion = "8.10.2"
     distributionUrl = "https://services.gradle.org/distributions/gradle-$gradleVersion-bin.zip"
     distributionSha256Sum = "31c55713e40233a8303827ceb42ca48a47267a0ad4bab9177123121e71524c26"
-}
\ No newline at end of file
+}
```
kropp, line 19: Please leave the linefeed at the end of file, to make this and following diffs cleaner

gradle/libs.versions.toml
```diff
@@ -94,7 +95,10 @@ android-svg = { module = "com.caverock:androidsvg-aar", version.ref = "android-s
 coil-compose = { module = "io.coil-kt.coil3:coil-compose", version.ref = "coil" }
 coil-network-ktor3 = { module = "io.coil-kt.coil3:coil-network-ktor3", version.ref = "coil" }
 
+aboutlibraries-compose = { module = "com.mikepenz:aboutlibraries-compose", version.ref = "aboutlibraries" }
+
```
kropp, line 99: Licenses list is pretty simple IMHO, and can be easily implemented in the app without 3rd party libraries, to better adhere to our design principles.
Can we use the library only to collect license information, but render it ourselves?

kropp:
The font looks different from the rest of the app, it's  definitely not JetBrains Sans.

kropp (changes requested):

gradle/libs.versions.toml
zsmb13: Yes! It can give us just the data itself and then we can render it arbitrarily.

e5l (approved):
Please consider using the same style as privacy policy screens

zsmb13:
Update on this PR: I removed the UI from the external library, now we're only using it to generate the JSON data with libraries and then parse it into an object. The UI implementation is now all within the app, using our themes etc. I didn't ask for the separate design as this is really one of those screens that will be used very very rarely.

https://github.com/user-attachments/assets/7716454e-6c53-47e2-b6f2-f9a2e475b7e6

build.gradle.kts
zsmb13: The auto-formatting added the newline here when touching the file. Just to clarify, are you asking to keep this newline or remove it? I'd personally keep it.

kropp: I'm for having a newline at the end of each file. I probably misread the diff, it looked like the newline was removed.

kropp (approved):

PR #240, opened 2025-01-18
Set up top level navigation

### LINKED ISSUE #236: Set up bottom navigation bar in the app

zsmb13:
Fixes #236 

This PR updates the navigation setup of the app the following way:
- Most screens are kept at the top level as before, within the root nav controller
- The onboarding flow "start screens" are in their own little nested nav graph
   - For now, there's a fake Flow that produces a random true/false value for whether the onboarding has been completed, we'll hook this up to actual preference storage later. This means the onboarding flow is skipped 50% of the time.
- The new `HomeScreen` component serves the screen with 4-button bottom navigation between its destinations. This can't be a nested nav graph as there's also common UI that persists through the screens (the bottom navigation bar), so it's a destination that contains a second `NavHost` that navigates within the `HomeScreen` destinations.
   - The Compose version bumps are needed for this nesting to work, as the new versions include [this bugfix](https://github.com/JetBrains/compose-multiplatform-core/pull/1596)


androidApp/src/androidMain/kotlin/org/jetbrains/kotlinconf/android/MainActivity.kt
```diff
@@ -20,7 +25,9 @@ class MainActivity : ComponentActivity() {
         )
 
         setContent {
-            App(context)
+            Box(Modifier.windowInsetsPadding(WindowInsets.systemBars)) {
```
zsmb13, line 28: Adding some basic insets here just to make buttons clickable, we'll need to do proper insets work later

ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/MainNavigation.kt
```diff
@@ -61,29 +62,30 @@ private fun MainNavigationButton(
...
+        destinations.forEach { destination ->
             MainNavigationButton(
-                iconResource = it.icon,
-                contentDescription = it.label,
-                selected = it == currentDestination,
-                onClick = { onSelected(it) },
+                iconResource = destination.icon,
+                contentDescription = destination.label,
+                selected = destination == currentDestination,
+                onClick = { onSelect(destination) },
```
kropp, line 88: According to the design mockup, the active icon should not only be highlighted, but it is also filled variant instead of outlined.

androidApp/src/androidMain/kotlin/org/jetbrains/kotlinconf/android/MainActivity.kt
kropp: I pushed a commit to this branch to fix the insets, as they bothered me for some time already

kropp (changes requested):

gradle/libs.versions.toml
```diff
@@ -13,8 +13,8 @@ androidx-preference = "1.2.1"
 
 kotlin = "2.1.0-firework.32" # custom version for experimental hot reload support
 
-compose-android = "1.7.1"
-compose-multiplatform = "1.7.0"
+compose-android = "1.7.6"
+compose-multiplatform = "1.8.0-alpha01"
```
kropp, line 17: `1.8.0-alpha02` has been published today

ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/MainNavigation.kt
zsmb13: Great catch! Fixed.

kropp (approved):

PR #244, opened 2025-01-23
App Terms of use and Privacy Policy screens

kropp:
They are separate and different from Visitor's Terms of Use and Privacy Policy

### DIFF
--- a/shared/src/commonMain/composeResources/values/strings.xml
+++ b/shared/src/commonMain/composeResources/values/strings.xml
@@ -86,6 +86,12 @@ We will process your data in accordance with the App Privacy Policy. You can adj
 
     <string name="privacy_policy_for_visitors">Privacy Policy</string>
     <string name="privacy_policy_for_visitors_title">KotlinConf 2024 Privacy Policy for Visitors</string>
-    <string name="privacy_policy_for_visitors_version">Version 1.0 of May 1, 2024</string>
+    <string name="privacy_policy_for_visitors_version">Version 1.1 of February 28, 2024</string>
+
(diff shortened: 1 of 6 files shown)


shared/src/commonMain/composeResources/values/strings.xml
```diff
@@ -86,6 +86,12 @@ We will process your data in accordance with the App Privacy Policy. You can adj
...
     <string name="privacy_policy_for_visitors">Privacy Policy</string>
     <string name="privacy_policy_for_visitors_title">KotlinConf 2024 Privacy Policy for Visitors</string>
-    <string name="privacy_policy_for_visitors_version">Version 1.0 of May 1, 2024</string>
+    <string name="privacy_policy_for_visitors_version">Version 1.1 of February 28, 2024</string>
+
+    <string name="app_privacy_policy_title">Privacy Policy</string>
+
+    <string name="app_terms">Terms of Use</string>
+    <string name="app_terms_title">KotlinConf App Terms of Use</string>
+    <string name="app_terms_version">Version 1.0 of May 1, 2024</string>
```
zsmb13, line 95: Are these 2024 dates intentional, are the docs the same as last year? Or will these will be updated later?

zsmb13 (approved):
Code looks good, just one question

shared/src/commonMain/composeResources/values/strings.xml
kropp: I don't have any new docs, but I've already logged a task to review all the legal texts.
Right now these are from the previous year.

PR #243, opened 2025-01-23
implement Speakers screen

kropp:
fixes #242

PR #247, opened 2025-01-26
Implement settings screen

zsmb13:
Fixes #243 

Wasn't in the design or anything, but added a splash of fun to the theme switcher:

https://github.com/user-attachments/assets/167bc04d-24be-4f2d-a301-3f3b4c360679


shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/ConferenceService.kt
```diff
@@ -34,6 +34,9 @@ class ConferenceService(
     private var userId2024: String? by storage.bind(String.serializer().nullable) { null }
     private var needsOnboarding: Boolean by storage.bind(Boolean.serializer()) { true }
     private var notificationsAllowed: Boolean by storage.bind(Boolean.serializer()) { false }
+    private var _themeStorage: Theme by storage.bind(Theme.serializer()) { Theme.SYSTEM }
+    private val _theme = MutableStateFlow(_themeStorage)
```
zsmb13, line 38: These can be consolidated later if we set up DataStore for storage, which always exposes the values as Flows anyway

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/App.kt
```diff
@@ -1,13 +1,16 @@
 package org.jetbrains.kotlinconf
 
+import URLs
```
e5l, line 3: would be nice to move it under our package

e5l (approved):
Looks good! Left minor comment

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/NotificationSettings.kt
```diff
@@ -0,0 +1,63 @@
...
+    var jetbrainsNews by mutableStateOf(true)
+
+    val model: NotificationSettings
+        @Composable
+        get() = NotificationSettings(
+            scheduleUpdates = scheduleUpdates,
+            kotlinConfNews = kotlinConfNews,
+            jetbrainsNews = jetbrainsNews,
+        )
+}
```
kropp, line 34: I have mixed feelings about duplicating settings with `mutableStateOf` per each setting. It may slightly improve performance by reducing the number of (re-)compositions, but I doubt it would be noticeable at all.
At the same time, it leads to code duplication and increases complexity.
I'd say passing a `NotificationsSettings` DTO directly would suffice.

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/SettingsScreen.kt
```diff
@@ -0,0 +1,244 @@
...
+        }
+        Spacer(Modifier.height(8.dp))
+        StyledText(
+            text = stringResource(
+                when (theme) {
+                    Theme.SYSTEM -> Res.string.settings_theme_system
+                    Theme.LIGHT -> Res.string.settings_theme_light
+                    Theme.DARK -> Res.string.settings_theme_dark
+                }
+            ),
```
kropp, line 238: I'd provide respective resources as parameters directly to this Composable.
It would increase the number of parameters just by one but will be more straightforward.
At a call site, replacing loop with three calls wouldn't make the code any worse as well.

kropp (approved):

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/App.kt
zsmb13: Good catch!

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/SettingsScreen.kt
zsmb13: For me it's cleaner if the `ThemeBox` can be used with just the `Theme` value passed in as a parameter, I'll leave it as-is for now.

PR #248, opened 2025-01-29
Create Schedule screen (part 1)

zsmb13:
Adds the initial implementation for the schedule screen. First PR of multiple, so that we don't need to review all of this in a single go.

Includes:
- A ViewModel for the sessions screen, offloading the list calculation operations to background threads
- Displaying the list of sessions
- Session feedback handling
- The top Day buttons skipping to the appropriate days (and updating based on scroll state)
- Filtering for bookmarked only
- Entering search mode, filtering for names and session titles, with handling of diacritics
- In search mode, filtering by tags

TODOs remaining (also in the code):
- The "Now" button for the header
- Displaying when a session is coming up soon or currently live
- Service events
- Some improvements on date and string handling (currently too hardcoded)

https://github.com/user-attachments/assets/2392fcb2-0da0-499d-9d08-d800e585d1d4


shared/store.json
kropp, line 1: I believe this file shouldn't be stored in the repo, we may exclude it via .gitignore

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/ScheduleScreen.kt
```diff
@@ -0,0 +1,289 @@
...
+                title = stringResource(Res.string.nav_destination_schedule),
+                startContent = {
+                    // TODO Now Button
+                },
+                endContent = {
+                    TopMenuButton(
+                        icon = UiRes.drawable.bookmark_24,
+                        selected = bookmarkFilterEnabled,
+                        onToggle = { onBookmarkFilter(it) },
+                        contentDescription = "Filter bookmarked"
```
kropp, line 188: String should be extracted to resources

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/ScheduleScreen.kt
```diff
@@ -0,0 +1,289 @@
...
+                    )
+                    TopMenuButton(
+                        icon = UiRes.drawable.search_24,
+                        onClick = { onHeaderStateChange(MainHeaderContainerState.Search) },
+                        contentDescription = "Search"
```
kropp, line 193: String should be extracted to resources

kropp (approved):
I added some minor comments, and there are a bunch of TODOs in the code, but given that it is just the first part of a bigger PR, I assume it is okay.

shared/store.json
zsmb13: Added to gitignore

PR #250, opened 2025-01-30
Set default params for desktop apps

zsmb13:


shared/src/jvmMain/kotlin/org/jetbrains/kotlinconf/main.kt
```diff
@@ -1,12 +1,16 @@
...
+import androidx.compose.ui.unit.dp
 import androidx.compose.ui.window.Window
 import androidx.compose.ui.window.application
+import androidx.compose.ui.window.rememberWindowState
 
 fun main() = application {
     Window(
         onCloseRequest = ::exitApplication,
         title = "KotlinProject",
+        alwaysOnTop = true,
```
kropp, line 12: I'd personally avoid it even for development purposes, but I don't mind leaving it like this.

kropp (approved):

shared/src/jvmMain/kotlin/org/jetbrains/kotlinconf/main.kt
zsmb13: We can remove it once Compose Hot Reload's UI respects the setting, but right now that UI is always on top anyway.

kropp: Yeah, this bothers me too.

PR #255, opened 2025-02-03
Use multiplatform-settings for storage

### LINKED ISSUE #256: Replace ApplicationStorage with DataStore or multiplatform-settings

zsmb13:
Fixes #256 

Converts the `ApplicationStorage` interface so that the setters are suspending methods and the getters return observable Flows of the stored values. This prepares us for potentially swapping the underlying storage to [DataStore](https://developer.android.com/topic/libraries/architecture/datastore) or similar libraries which are all coroutine-based. However, DataStore doesn't support the wasm target yet.

Given the above, this PR introduces the [multiplatform-settings](https://github.com/russhwolf/multiplatform-settings) library for the implementation of storage. It uses the following implementations for the platforms:
- SharedPrefs for Android
- NSUserDefaults for iOS
- Properties file for desktop JVM (the `Preferences`-based implementation unfortunately can't store long values)
- localstorage for web

It would be nice to have a real database implementation (like Room) for storing the list of sessions, but that doesn't work for us on the web target, so we're sticking to just a key-value store instead.

---

As the API of `ApplicationStorage` was modified, this is a good chance to rewrite the API surface of `ConferenceService` accordingly. Many methods have been changed to be suspending / return Flows, and most storage directly inside `ConferenceService` was eliminated in favour of propagating Flows from the storage implementation.

The single-use methods called from `ConferenceService#init` have been inlined and simplified as well.


shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/ConferenceService.kt
```diff
@@ -50,77 +41,76 @@ val UNKNOWN_SPEAKER: Speaker = Speaker(
...
+) {
+    private val storage: ApplicationStorage = MultiplatformSettingsStorage(context)
 
     private val client: APIClient by lazy {
         APIClient(endpoint)
     }
 
-    override val coroutineContext: CoroutineContext =
-        SupervisorJob() + Dispatchers.App
+    val scope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
```
e5l, line 51: Could you tell me why do we need to have explicit scope?

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/ScheduleViewModel.kt
```diff
@@ -251,8 +251,7 @@ class ScheduleViewModel(
 
     fun onBookmark(sessionId: SessionId, bookmarked: Boolean) {
         viewModelScope.launch {
-            // TODO would be nicer if we could pass a boolean in here
-            service.toggleFavorite(sessionId)
+            service.toggleFavorite(sessionId, bookmarked)
```
e5l, line 254: toggle -> set

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/storage/MultiplatformSettingsStorage.kt
```diff
@@ -0,0 +1,43 @@
...
+import org.jetbrains.kotlinconf.SessionId
+import org.jetbrains.kotlinconf.Theme
+
+expect fun createSettings(context: ApplicationContext): ObservableSettings
+
+@OptIn(ExperimentalSettingsApi::class)
+class MultiplatformSettingsStorage(context: ApplicationContext) : ApplicationStorage {
+    private val settings = createSettings(context)
+
+    override fun getUserId(): Flow<String?> = settings.getStringOrNullFlow("userid2025")
```
e5l, line 23: Should we extract string keys to avoid typos?

e5l (approved):
lgtm, left some comments

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/ConferenceService.kt
zsmb13: It's what the recommendation is since https://github.com/Kotlin/kotlinx.coroutines/issues/1581

For me the main benefit is that the scope can be private within the class (which it wasn't here, but fixed in a new commit). It's weird if callers can externally do `service.launch {}` to run something inside the service's scope.

It also allows for the scope to be injected for testing, which we're not doing here, at least not yet, but it gets us one step closer to that.

PR #257, opened 2025-02-04
Rework order of bottom navigation entries

zsmb13:
Moves the Info screen to be the last item of the bottom navigation bar


e5l (approved):
lgtm

PR #258, opened 2025-02-05
Create Schedule screen (part 2)

zsmb13:
Adds the Now button to the header of the screen. Time for the service is now provided by a `TimeProvider`, which has a fake implementation that can put our current time in the middle of the test data from 2024 that we're currently working with.

https://github.com/user-attachments/assets/30e14e31-f26b-4b3e-ae02-42eef67eec20

TODOs remaining on this screen:

- Displaying when a session is coming up soon
- Service events
- Improvements on date and string handling (currently too hardcoded)

### DIFF
--- a/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/App.kt
+++ b/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/App.kt
@@ -63,8 +63,8 @@ private fun koinConfiguration(context: ApplicationContext) = koinConfiguration {
         single { APIClient(URLs.API_ENDPOINT) }
         single<ApplicationStorage> { MultiplatformSettingsStorage(context) }
         single { NotificationManager(context) }
-//        single<TimeProvider> { ServerBasedTimeProvider(get()) }
-        single<TimeProvider> { FakeTimeProvider() }
+        single<TimeProvider> { ServerBasedTimeProvider(get()) }
+//        single<TimeProvider> { FakeTimeProvider() }
         singleOf(::ConferenceService)
     }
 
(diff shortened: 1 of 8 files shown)


shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/App.kt
```diff
@@ -57,13 +60,12 @@ fun App(context: ApplicationContext) {
...
-        single<ConferenceService> {
-            ConferenceService(
-                get<ApplicationContext>(),
-                URLs.API_ENDPOINT
-            )
-        }
+        single { APIClient(URLs.API_ENDPOINT) }
+        single<ApplicationStorage> { MultiplatformSettingsStorage(context) }
+        single { NotificationManager(context) }
+        single<TimeProvider> { ServerBasedTimeProvider(get()) }
```
e5l, line 66: nice idea

e5l (approved):
lgtm

kropp:
Could you please explain the reasoning for a custom `TimeProvider`? `kotlinx-datetime` features `Clock` interface suitable for injection. It lacks `StateFlow`, but it can be added as an extension.

zsmb13:
@kropp We're currently not using `kotlinx-datetime` types in the app. I created an issue (https://github.com/JetBrains/kotlinconf-app/issues/260) to move to using them. If we do that, we can probably replace this with `Clock`.

Although I'd also like to improve this fake implementation in later PRs for testing more scenarios. It could start from a hardcoded value but actually tick upwards at real time (or faster than real time) pace, to let us see the values of the schedule updating as time passes.

PR #264, opened 2025-02-06
Prompt for privacy policy screen if not yet accepted

### LINKED ISSUE #263: Prompt for privacy policy if not yet accepted whenever the user attempts to vote

zsmb13:
Fixes #263 

https://github.com/user-attachments/assets/8d1e1b8c-090d-4a91-927b-24180d359e64


shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/StartNotificationsScreen.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.jetbrains.kotlinconf.ui.components.Button
+import org.jetbrains.kotlinconf.ui.components.StyledText
+import org.jetbrains.kotlinconf.ui.theme.KotlinConfTheme
+import kotlinconfapp.shared.generated.resources.Res as AppRes
+
+@Composable
+fun StartNotificationsScreen(
+    onDone: (NotificationSettings) -> Unit,
+) {
+    // TODO populate with real values
```
e5l, line 37: Please consider logging an issue. Maybe we can use DI for it right away.

e5l (approved):
lgtm

kropp:
Judging from the video, if user hasn't accepted privacy policy and press reaction button, then, after they accept the policy and get back to the previous screen, the reaction is not selected and it should be clicked again.

Feel free to extract it to a separate issue.
Otherwise, LGTM

kropp (approved):

zsmb13:
@kropp Indeed, we don't have that implemented yet. It was simpler to just throw away the action in these cases, but it would be nicer if we could preserve the pending vote/comment and submitted it. Created https://github.com/JetBrains/kotlinconf-app/issues/267

Issue #269, opened 2025-02-10
Implement service items


zsmb13:
We need a way for the API to provide the times of service events (registration, lunch, coffee, party, etc.) and add these to the list of schedule items.

PR #270, opened 2025-02-10
Create Schedule screen (part 3)

zsmb13:
Adds the following:

- Displaying when a session is coming up soon
- Service events (except #269 which will add actual data)

TODOs remaining on this screen:

- #269 
- Improvements on date and string handling (currently too hardcoded)

https://github.com/user-attachments/assets/2820a91d-7e7a-4b6d-9e4b-74642d3b2f43

### DIFF
--- a/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/TimeProvider.kt
+++ b/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/TimeProvider.kt
@@ -3,10 +3,14 @@ package org.jetbrains.kotlinconf
 import io.ktor.util.date.GMTDate
 import io.ktor.util.date.Month
 import io.ktor.util.date.plus
+import kotlinx.coroutines.awaitCancellation
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.update
+import kotlin.time.Duration.Companion.minutes
+import kotlin.time.Duration.Companion.seconds
 
 interface TimeProvider {
     fun now(): GMTDate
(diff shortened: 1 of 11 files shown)


shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/TimeProvider.kt
```diff
@@ -38,15 +42,27 @@ class ServerBasedTimeProvider(private val client: APIClient) : TimeProvider {
...
+    override val time: StateFlow<GMTDate> = _time
+    override fun now(): GMTDate = baseTime
     override suspend fun run(): Nothing {
-        while (true) {
-            delay(Long.MAX_VALUE) // This waits forever on its own
+        if (freezeTime) {
+            awaitCancellation()
+        } else {
+            while (true) {
+                // Progress time at 4x speed for testing
```
e5l, line 58: Please log an issue or at least add `TODO`, so we don't forget to drop it on review

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/TimeProvider.kt
```diff
@@ -38,15 +42,27 @@ class ServerBasedTimeProvider(private val client: APIClient) : TimeProvider {
...
+                delay(15.seconds)
+                _time.update { t ->
+                    t.plus(1.minutes).also {
+                        println("Fake time is now $it")
```
e5l, line 62: should we use logger?

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/ScheduleScreen.kt
```diff
@@ -305,39 +315,105 @@ fun ScheduleList(
...
+        lightning = session.isLightning,
+        time = session.badgeTimeLine,
+        timeNote = session.startsInMinutes?.let { count ->
+            stringResource(Res.string.schedule_in_x_minutes, count)
+        },
+        status = when {
+            session.isFinished -> TalkStatus.Past
+            session.isLive -> TalkStatus.Now
+            session.isUpcoming -> TalkStatus.Upcoming
+            else -> TalkStatus.Upcoming // Shouldn't happen
```
e5l, line 408: It looks like we need to log an error here

zsmb13: Good note - replaced this whole mess with an enum so we always have a specific session state.

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/TimeProvider.kt
zsmb13: This is meant to stay like this, as this `FakeTimeProvider` is unused unless you modify the DI setup in `App.kt`. But if you choose to use it, it'll progress time faster so that it's easier to see how the schedule updates.

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/TimeProvider.kt
zsmb13: Since this is only for development purposes, I don't think it's worth it

PR #272, opened 2025-02-11
wip kotlinx.datetime, please ignore

zsmb13:


shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/SessionCardView.kt
```diff
@@ -1,40 +1,36 @@
...
+    val isLightning: Boolean = endsAt - startsAt <= LIGHTNING_TALK_LIMIT
+}
 
 val SessionCardView.isLive get() = state == SessionState.Live
 val SessionCardView.isUpcoming get() = state == SessionState.Upcoming
 val SessionCardView.isPast get() = state == SessionState.Past
 
-val Session.isLightning: Boolean get() = endsAt.timestamp - startsAt.timestamp <= 15 * 60 * 1000
+val Session.isLightning: Boolean
+    get() = endsAt - startsAt <= LIGHTNING_TALK_LIMIT
```
kropp, line 33: minor: `isLightning` check is duplicated

backend/src/main/kotlin/org/jetbrains/kotlinconf/backend/Api.kt
```diff
@@ -1,15 +1,22 @@
...
+import io.ktor.server.routing.route
+import io.ktor.util.date.GMTDate
+import kotlinx.datetime.toInstant
+import org.jetbrains.kotlinconf.EVENT_TIME_ZONE
+import org.jetbrains.kotlinconf.FeedbackInfo
+import org.jetbrains.kotlinconf.VoteInfo
 import org.jetbrains.kotlinconf.Votes
-import java.time.*
+import java.time.Clock
+import java.time.LocalDateTime
```
kropp, line 19: Is `java.time` still needed here?

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/SessionCardView.kt
zsmb13: We need this both on `Session` and `SessionCardView`, I think with the limit variable extracted at least it's alright like this

backend/src/main/kotlin/org/jetbrains/kotlinconf/backend/Api.kt
zsmb13: The backend code still uses `java.time` APIs in a bunch of places, I tried touching as little of the backend code as possible. Only edited the parts that were tightly coupled to the shared code.

We *could* rewrite the backend to be all `kotlinx.datetime` too, but it's not strictly necessary. Just wanted to have these types on the client side instead of using the Ktor `GMTDate` type everywhere.

@e5l to review the backend changes though, and whether we want more of them.

e5l: Feel free to use kotlinx.datetime if it's possible

e5l (approved):
lgtm

kropp (approved):

PR #274, opened 2025-02-11
Update notification and permission handling

zsmb13:
Updates the notification system (and related permission handling) to some new APIs, still implemented with separate native APIs on each platform, as no suitable KMP library was found that covered our use cases (including scheduled notifications).

Makes notification handling depend on `TimeProvider` as well, allowing tests of notifications with fake time values and sped up clocks within the app.


shared/src/iosMain/kotlin/org/jetbrains/kotlinconf/IOSNotificationService.kt
```diff
@@ -0,0 +1,72 @@
...
+                    this.calendar = calendar
+                },
+                repeats = false,
+            )
+        } else {
+            null
+        }
+        val request = UNNotificationRequest.requestWithIdentifier(notificationId, content, trigger)
+        notificationCenter.addNotificationRequest(request) { error: NSError? ->
+            // TODO use logger
```
zsmb13, line 63: This is a TODO for all the `println` calls added here and later. Please ignore these for now, I created https://github.com/JetBrains/kotlinconf-app/issues/285 to fix this in a follow-up PR (to not add even more changes within this one).

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/ScheduleScreen.kt
```diff
@@ -138,6 +138,12 @@ fun ScheduleScreen(
         viewModel.setSearchParams(params)
     }
 
+    if (items.isNotEmpty() && firstLiveIndex != -1) {
+        LaunchedEffect(Unit) {
+            listState.scrollToItem(firstLiveIndex)
+        }
+    }
```
zsmb13, line 145: Quick random fix: scroll to current time in schedule when first loaded

shared/src/iosMain/kotlin/org/jetbrains/kotlinconf/IOSNotificationService.kt
zsmb13: That PR is already here: https://github.com/JetBrains/kotlinconf-app/pull/286

shared/src/wasmJsMain/kotlin/org/jetbrains/kotlinconf/main.wasm.kt
```diff
@@ -2,12 +2,28 @@ package org.jetbrains.kotlinconf
...
+external object JsJodaTimeZoneModule
+
+private val jsJodaTz = JsJodaTimeZoneModule
 
 @OptIn(ExperimentalComposeUiApi::class)
 fun main() {
     initCoil()
     CanvasBasedWindow {
-        App(ApplicationContext())
+        App(platformModule = module {
```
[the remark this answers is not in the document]
zsmb13: Doesn't work here as we're also calling the `module` function, but I'd also avoid it either way because it would look confusing in Compose code

shared/src/webMain/kotlin/org/jetbrains/kotlinconf/ServiceWorkerNotificationService.kt
```diff
@@ -0,0 +1,76 @@
...
+}
+
+private fun registerNotificationByServiceWorker(
+    delay: Long,
+    notificationId: String,
+    title: String,
+    message: String
+): Unit =
+    js(
+        """{
```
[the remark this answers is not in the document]
zsmb13: I'll see if this can be moved to the JS side easily, but this is something I moved over from the old implementation as-is. I'd also be happy to just leave it and not touch it, since it works :D

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/NotificationService.kt
```diff
@@ -0,0 +1,16 @@
+package org.jetbrains.kotlinconf
+
+import kotlinx.datetime.LocalDateTime
+
+interface NotificationService {
+    suspend fun requestPermission(): Boolean
+
+    fun post(
+        notificationId: String,
```
kropp, line 9: It would be nice to use `value class NotificationId` here, similar to other places.

shared/src/androidMain/kotlin/org/jetbrains/kotlinconf/AndroidNotificationService.kt
```diff
@@ -0,0 +1,190 @@
...
+import androidx.core.content.getSystemService
+import kotlinx.coroutines.channels.Channel
+import kotlinx.datetime.LocalDateTime
+import kotlinx.datetime.toInstant
+
+private const val EXTRA_TITLE = "title"
+private const val EXTRA_MESSAGE = "message"
+private const val EXTRA_NOTIFICATION_ID = "notificationId"
+private const val EXTRA_ICON_ID = "iconId"
+private const val NOTIFICATION_CHANNEL_ID = "channel_all_notifications"
```
kropp, line 25: Should we introduce separate channels for different notification types (session alerts vs news)?

kropp (approved):

shared/src/androidMain/kotlin/org/jetbrains/kotlinconf/AndroidNotificationService.kt
zsmb13: Not without putting in a lot more work for managing these channels. The ideal setup for Android users would probably be to use 3 different notification channels and not have in-app settings for notification categories at all, only a button that sends them to the system settings where they can configure each channel. But it's a bunch of extra work just for the one platform, so not critical for the time being.

(We can't read or change the state of notification channels from the app, so we can't sync our own toggles to what the user sets on the channels in settings.)

zsmb13: On the bright side, we'll bump the minimum API to 26 on the icons PR, which is where notification channels were introduced, so that eliminates a whole problem where on 24-25 we'd have to have the in-app switcher instead anyway.

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/NotificationService.kt
zsmb13: We need to serialize it into an intent at some point on Android, and pass it to JS for interop, I don't want to complicate its type for now

PR #275, opened 2025-02-12
Update Compose Multiplatform and Compose Hot Reload versions

zsmb13:


kropp (changes requested):
Navigation library has been updated as well, let's promote it too

kropp:
I tried it and it now crashes on start on iOS, due to a [known issue](https://youtrack.jetbrains.com/issue/CMP-7571)
`kotlin.native.internal.IrLinkageError: Can not read value from backing field of property 'androidx_compose_runtime_ProvidedValue$stable': Private backing field of property declared in module <org.jetbrains.compose.runtime:runtime> can not be accessed in module <io.insert-koin:koin-compose>`

I'd suggest to postpone update for a bit

zsmb13:
@kropp We already have a crash on the current version when trying to load images with Coil on native, so I think we might as well bump the version.

```
Uncaught Kotlin exception: kotlin.native.internal.IrLinkageError: Function 'layout' can not be called: No function found for symbol 'androidx.compose.ui.layout/MeasureScope.layout|layout(kotlin.Int;kotlin.Int;kotlin.collections.Map<androidx.compose.ui.layout.AlignmentLine,kotlin.Int>;kotlin.Function1<androidx.compose.ui.layout.Placeable.PlacementScope,kotlin.Unit>){}[0]'
```

https://github.com/coil-kt/coil/issues/2771

(This issue is actually fixed in alpha03 😄)

gradle/libs.versions.toml
```diff
@@ -15,8 +15,8 @@ androidx-preference = "1.2.1"
 kotlin = "2.1.10-firework.33"
 
 compose-android = "1.7.6"
```
MatkovIvan, line 17: Please keep android and multiplatform versions in sync

```suggestion
compose-android = "1.8.0-beta01"
```

zsmb13: Thanks for catching that!

PR #283, opened 2025-02-19
Merge workshop parts in a single one

### LINKED ISSUE #259: Merge workshop events into one per workshop
Sessionize contains the workshops in several pieces for each workshop
```
Some Workshop, Part 1 (9:00-10:30)
Some Workshop, Part 2 (11:00-12:30)
Some Workshop, Part 3 (13:30-15:00)
Some Workshop, Part 4 (15:30-17:00)
```

We'd like to receive this in the app as a single large session to make things easier to display:
```
Some Workshop (9:00-17:00)
```

e5l:
Fix #259


backend/src/main/kotlin/org/jetbrains/kotlinconf/backend/Sessionize.kt
```diff
@@ -86,4 +89,35 @@ fun SessionizeData.toConference(): Conference {
...
+            startTime,
+            endTime,
+            first.tags
+        )
+    }
+
+    return nonWorkshop + workshops
+}
+
+internal fun String.dropAfter(chunk: String): String {
```
zsmb13, line 120: I think this is https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/substring-before.html

backend/src/main/kotlin/org/jetbrains/kotlinconf/backend/Sessionize.kt
```diff
@@ -8,14 +8,15 @@ import io.ktor.server.application.log
...
+import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.launch
 import org.jetbrains.kotlinconf.Conference
 import org.jetbrains.kotlinconf.Session
 import org.jetbrains.kotlinconf.Speaker
 import java.util.concurrent.TimeUnit
 
-@Volatile
-private var conference: Conference? = null
+private val conference = MutableSharedFlow<Conference>()
```
zsmb13, line 19: Shouldn't this be a `MutableStateFlow` so it always holds the latest value that was stored in it? With that we could make `getSessionizeData` non-suspending again, and just return `conference.value`.

zsmb13 (changes requested):
Small notes

backend/src/main/kotlin/org/jetbrains/kotlinconf/backend/Sessionize.kt
e5l: Thank you, will fix

backend/src/main/kotlin/org/jetbrains/kotlinconf/backend/Sessionize.kt
e5l: To get the initial state we need to suspend, so I'm not sure if it is possible to replace it

zsmb13: Makes sense! I didn't look at the usage in detail before, but now I see that the API call to grab the data needs to wait for this. In this case, we should probably have a `replay = 1` set on this, to make sure that once a value is set, it's kept around?

e5l: done, added test

zsmb13 (approved):

PR #289, opened 2025-02-27
Change icon for the 2025 edition

kropp:


zsmb13:
The new Android icon looks just a tiny bit too large to me:
image

Compare with the size of the monochrome icon, which I think is perfect:
image

gradle/libs.versions.toml
```diff
@@ -2,7 +2,7 @@
 aboutlibraries = "11.4.0"
 agp = "8.7.3"
 android-compileSdk = "35"
-android-minSdk = "24"
+android-minSdk = "26"
```
zsmb13, line 5: Bumped to support `adaptive-icon`.

e5l: lgtm

e5l:
The icon looks weird on the white background. Let's ask the designer if we can do it better

e5l (approved):

PR #291, opened 2025-02-28
Navigate to session page when notification is clicked

zsmb13:
Android:

https://github.com/user-attachments/assets/9f0d09a0-2c98-4db8-abd0-ffbea529d6d3

iOS:

https://github.com/user-attachments/assets/0ba65e76-1789-4a74-a133-f81f59b31982


shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/navigation/KotlinConfNavHost.kt
```diff
@@ -43,10 +45,24 @@ import org.jetbrains.kotlinconf.screens.TermsOfUse
 import org.jetbrains.kotlinconf.utils.getStoreUrl
 import kotlin.reflect.typeOf
 
+fun navigateToSession(notificationId: String) {
+    val sessionId = SessionId(notificationId.substringBefore("-"))
+    sessionIds.trySend(sessionId)
+}
```
zsmb13, line 51: We can make this more sophisticated later, especially if we need to link to other types of pages, like news

e5l (approved):
lgtm

kropp (approved):

PR #297, opened 2025-03-04
Update sessionize url to 2025

e5l:
While the server is not deployed, the local server should be run.
You can do this by calling:
```
./gradlew :backend:run
```
before starting the android application


kropp (approved):

zsmb13 (approved):

zsmb13:
Note that for desktop testing the local URL needs to be `0.0.0.0:8080`. Given that we'll remove that part anyway though, it's fine to use the Android address - that's the more difficult one to figure out anyway.

e5l:
I'll figure out the staging before merging

PR #299, opened 2025-03-05
Screen backgrounds

kropp:
Explicitly set screen backgrounds, otherwise they remain transparent and other screens are shining through during navigation transitions and other screens are shining through during navigation transitions.


zsmb13 (approved):
Ready to go after we merge the version bumps that this depends on

e5l (approved):

zsmb13:
I think it's good to go.

There's a slightly ugly workaround with calling `NavHost` twice depending on a condition, and I'm not 100% sure of the animation used yet for Android, but I think they're decent. If you can check the Android animations and it looks fine on your end, we can merge it.

kropp:
LGTM

PR #305, opened 2025-03-06
Add storage versioning

zsmb13:
Keep track of current storage version. If we see an older storage version (uninitialized at the moment means it's from before 2025), we destructively migrate to the new version.

Later on, as the storage schema changes, we can create more sophisticated migrations based on the stored and current storage versions.

Fixes [#306](https://github.com/JetBrains/kotlinconf-app/issues/306)


kropp (approved):

e5l (approved):
lgtm. We may ask legal if we should ask the privacy policy and notification permissions every year

zsmb13:
Easiest to ask every year for now, just in case the text changes. Doesn't really hurt the user that much to accept it once a year even if it happens to be the same 😄

PR #309, opened 2025-03-07
Implement Firebase push notifications on Android and iOS

zsmb13:
Add Firebase push notification handling for Android and iOS, using the [KMPNotifier](https://github.com/mirzemehdi/KMPNotifier) library.

The app listens to three topics (controlled by the in-app settings):

* `ScheduleUpdates`
* `KotlinConfNews`
* `JetBrainsNews`

In addition to the title and message which will be displayed in the notification, the pushes may contain custom data with the following keys to trigger navigation when the notification is opened:

* `newsId`
* `sessionId`

These constants are also found in `PushNotificationConstants` to reuse.


shared/src/jvmMain/kotlin/org/jetbrains/kotlinconf/NotImplementedLocalNotificationService.kt
```diff
@@ -0,0 +1,9 @@
+package org.jetbrains.kotlinconf
+
+import kotlinx.datetime.LocalDateTime
+
+class NotImplementedLocalNotificationService : LocalNotificationService {
```
e5l, line 5: I would name it Blank. Not implemented associates with exception for me

shared/src/iosMain/kotlin/org/jetbrains/kotlinconf/IOSNotifications.kt
```diff
@@ -0,0 +1,31 @@
...
+    val content: UNNotificationContent = response.notification.request.content
+    val notificationId = content.userInfo[IOSLocalNotificationService.LOCAL_NOTIFICATION_ID_KEY] as? String
+    if (notificationId != null) {
+        // Local notification clicked
+        navigateByLocalNotificationId(notificationId)
+        return
+    }
+
+    // Process push notifications
+    // TODO simplify this to NotifierManager.onNotificationClicked(content) and inline notifierManagerListener
```
e5l, line 19: please log an issue

e5l (approved):

PR #311, opened 2025-03-07
Improve voting

kropp:
Some Friday evening fun: added haptic feedback to session voting

Currently, feedback forms are hidden, so you'll need to enable them manually in the code

@zsmb13 could you please check whether these haptics make sense on Android?


zsmb13:
Found a crash while testing this, so added some error handling improvements.

I think we can actually simplify this. Whichever option the user votes for, the interaction itself was successful, so I think it should trigger with `HapticFeedbackType.Confirm` in all cases, looking at its docs. We could also add that to the `SessionScreen` which has feedback on it.

kropp:
I'm ok with all haptics to be `HapticFeedbackType.Confirm`, just wanted to play a bit with it. Also, it will simplify the code.

PR #312, opened 2025-03-07
Move Koin init to app launch

zsmb13:
Moves Koin initialization outside the Compose `App()` function. This lets us get our dependency graph up and running without the UI necessarily being initialized, which will be required for cases like notification handling, when the app is started but the `MainActivity` (or equivalent) isn't.

This also lets us move the `DevelopmentEntryPoint` out to the JVM `main` function only.


androidApp/src/androidMain/AndroidManifest.xml
```diff
@@ -13,6 +13,7 @@
         android:icon="@drawable/ic_launcher"
         android:label="@string/app_name"
         android:supportsRtl="true"
+        android:name="org.jetbrains.kotlinconf.android.KotlinConfApplication"
```
e5l, line 16: Please make sure the name and location is matching app location for the previous year, so we will be able to publish an update

e5l (approved):
lgtm

androidApp/src/androidMain/AndroidManifest.xml
zsmb13: This here is only the reference to the application class used, it won't affect anything external-facing like the package name.

androidApp/src/androidMain/kotlin/org/jetbrains/kotlinconf/android/MainActivity.kt
```diff
@@ -27,14 +28,16 @@ class MainActivity : ComponentActivity() {
...
-            application = application,
-            notificationIconId = R.drawable.kotlinconf_notification_icon,
-        )
+        val permissionHandler = PermissionHandler(activity = this)
+        val activityModule = module {
+            single<PermissionHandler> { permissionHandler }
+        }
+        KoinPlatformTools.defaultContext()
+            .get()
+            .loadModules(listOf(activityModule))
```
arnaudgiuliani, line 37: instead of loading a side module here, you can have a definition with dynamic param like:
`factory { (activity : ComponentActivity) -> PermissionHandler(activity = activity) }`

When injecting, you pass the param. In MainActivity, you can have a field: `val permissionHandler : PermissionHandler by inject { parametersOf(this)}`. This will pass the activity as parameter.

shared/src/androidMain/kotlin/org/jetbrains/kotlinconf/AndroidNotificationService.kt
```diff
@@ -9,43 +9,42 @@ import android.content.Context
...
     private val alarmManager = context.getSystemService<AlarmManager>()
 
     override suspend fun requestPermission(): Boolean {
         if (Build.VERSION.SDK_INT < Build.VERSION_CODES.TIRAMISU) return true
 
         val permissions = arrayOf(Manifest.permission.POST_NOTIFICATIONS, Manifest.permission.USE_EXACT_ALARM)
-        permissionLauncher.launch(permissions)
-        return permissionResult.receive()
+
+        val permissionHandler = KoinPlatform.getKoin().getOrNull<PermissionHandler>()
```
arnaudgiuliani, line 46: You can use the `getKoin()` Service extension instead if you want.

zsmb13: Despite the name, this is not an `android.app.service`, so I don't think we can do that here. Does `getKoin()` perform the exact same `KoinPlatform.getKoin()` internally anyway?

androidApp/src/androidMain/kotlin/org/jetbrains/kotlinconf/android/MainActivity.kt
zsmb13: Unfortunately we inject the `PermissionHandler` in a deeper layer of the application, inside the `NotificationService` class. This is a data layer dependency of a `ViewModel`, and down there we have no way of accessing the `Activity` reference for injection.

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/Koin.kt
```diff
@@ -0,0 +1,49 @@
...
+            viewModelOf(::NewsListViewModel)
+            viewModelOf(::StartNotificationsViewModel)
+            viewModelOf(::NewsDetailViewModel)
+            viewModelOf(::SpeakersViewModel)
+            viewModelOf(::SpeakerDetailViewModel)
+        }
+
+        // Note that the order of modules here is significant, later
+        // modules can override dependencies from earlier modules
+        modules(platformModule, appModule, viewModelModule)
```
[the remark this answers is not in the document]
zsmb13: I just read the docs because I didn't hear about `include` before, but I don't fully understand what difference it would make in this case. Seems like it adds all these modules to the dependency graph just about the same way here.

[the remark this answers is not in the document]
arnaudgiuliani: The "includes" ensures that all modules are included efficiently and in order without collision.  The + operator you can have a risk of overloading the same definition.

androidApp/src/androidMain/kotlin/org/jetbrains/kotlinconf/android/MainActivity.kt
arnaudgiuliani: then in this case, you can declare on the fly your instance. It avoid you to have to define a module.
In the activity, you can do `getKoin().declare(PermissionHandler(activity = this))`

And later inject your PermissionHandler.

zsmb13: Neat, changed this to use `declare`!

androidApp/src/androidMain/kotlin/org/jetbrains/kotlinconf/android/MainActivity.kt
```diff
@@ -28,13 +27,8 @@ class MainActivity : ComponentActivity() {
...
         processIntent(intent)
 
-        val permissionHandler = PermissionHandler(activity = this)
-        val activityModule = module {
-            single<PermissionHandler> { permissionHandler }
-        }
-        KoinPlatformTools.defaultContext()
-            .get()
-            .loadModules(listOf(activityModule))
+        KoinPlatformTools.defaultContext().get()
```
arnaudgiuliani, line 30: `KoinPlatformTools.defaultContext().get()` can be shorter to `KoinPlatform.getKoin()`. It's the same

zsmb13: Oops, yes, I even used that elsewhere. Fixed!

PR #313, opened 2025-03-10
Workshop infinite scroller

zsmb13:
https://github.com/user-attachments/assets/7ed5807a-0025-4837-b42b-4694bec00899


shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/ScheduleScreen.kt
```diff
@@ -334,28 +335,32 @@ fun ScheduleList(
                 is WorkshopItem -> {
                     val workshops = item.workshops
                     Column(modifier = Modifier.fillMaxWidth()) {
-                        val pagerState = rememberPagerState(pageCount = { workshops.size })
+                        val pagerState = rememberPagerState(
+                            pageCount = { Int.MAX_VALUE },
+                            initialPage = Int.MAX_VALUE / 2,
```
e5l, line 340: looks weird, I would expect some comment

zsmb13: Fair enough, added comments :)

PR #321, opened 2025-03-14
Add support for pending user IDs

zsmb13:


shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/APIClient.kt
```diff
@@ -97,7 +95,7 @@ class APIClient(
      * Vote for session.
      */
     suspend fun vote(sessionId: SessionId, score: Score?): Boolean {
-        if (userId == null) return false
+        userId ?: return false
```
kropp, line 98: I personally find `if` check easier to read.
And also it works better with `if (userId != null) return true` in `ConferenceService::acceptPrivacyPolicy`, which cannot be shortened like this.

kropp (approved):

e5l (approved):

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/APIClient.kt
zsmb13: Fair enough :)

PR #322, opened 2025-03-17
Set dark/light mode values for resource loading

zsmb13:


terrakok:
Does it work, when you change the theme in the app? I mean does it load dark resources?

zsmb13:
@terrakok It does in my testing, yes. Look at the `by JetBrains` text in this drawable for example for the change:

https://github.com/user-attachments/assets/6aaf2798-85af-4421-8995-6cbe895594e9

terrakok:
It is not necessary but if you want to make it perfect... 😄

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/navigation/KotlinConfNavHost.kt
```diff
@@ -138,31 +138,31 @@ fun NavGraphBuilder.screens(navController: NavHostController) {
...
             onGeneralTerms = { navController.navigate(TermsOfUseScreen) },
             onWebsiteLink = { urlHandler.openUri(URLs.KOTLINCONF_HOMEPAGE) },
-            onBack = navController::popBackStack,
+            onBack = navController::navigateUp,
             onSpeaker = { speakerId -> navController.navigate(SpeakerDetailScreen(speakerId)) },
         )
     }
     composable<CodeOfConductScreen> {
-        CodeOfConduct(onBack = navController::popBackStack)
+        CodeOfConduct(onBack = navController::navigateUp)
```
terrakok, line 146: this change seems unrelated

terrakok (approved):

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/navigation/KotlinConfNavHost.kt
zsmb13: Yes, this part is a bit of a cheat, but I just noticed while going back-and-forth with theme settings a lot that I can pop everything off the backstack and end up with an empty frame, so I think it's time to make this change.

terrakok: an empty frame should be fixed already in the latest dev build

terrakok: `popBackStack` is the correct function. the `navigateUp` is required when you need to leave the app when the app was opened by a deeplink on the android, as I know

e5l (approved):

PR #330, opened 2025-03-18
Only scroll to current time on schedule screen once

zsmb13:


kropp (approved):

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/ScheduleScreen.kt
```diff
@@ -114,9 +114,13 @@ fun ScheduleScreen(
     }
 
     // Scroll to first live event on first content load
-    LaunchedEffect(state is ScheduleUiState.Content) {
-        if (state is ScheduleUiState.Content && state.firstLiveIndex != -1) {
-            listState.scrollToItem(state.firstLiveIndex)
+    var firstScrollPerformed by rememberSaveable { mutableStateOf(false) }
```
e5l, line 117: should we scroll again if the menu clicked second time?

e5l (approved):

PR #332, opened 2025-03-18
Partners screen update

kropp:


kropp:
Looks like it is going to conflict with #322, will update as soon as it is merged

PR #333, opened 2025-03-18
Only use the description of the first workshop part when merging workshops

zsmb13:
Currently we're creating workshop sessions that include the description 4 times in a row


e5l (approved):
lgtm

PR #335, opened 2025-03-20
Gradle Suggestions: Explicit Artifact Repository declarations & usage of hierarchy templates

sellmair:
Feel free to pick from this PR whatever feels right to you. 
Commit 1 is supposed to help Gradle with initial sync times by avoiding many unnecessary network requests. 

Tested: Clean sync (1 Gbit network, only one attempt to measure, your milage might vary)
master: 9m 48s
after:    6m 55s

Commit 2 is supposed to replace the discouraged usage of 'dependsOn' in source sets


zsmb13 (approved):
Brilliant, thanks for the contribution!

I made the repo setup a bit more flexible so that we don't have to explicitly list plugins coming from the Gradle plugin portal (we tend to grab some third-party things from there, could easily get annoying to update).

Also simplified the source set / dependency code a bit now that we no longer have to wire source sets with `dependsOn`.

PR #342, opened 2025-03-21
Bind compose navigation with browser history

eymar:


bashor:
@eymar 
Some questions/suggestions for future improvements:
* Can we pop history when a user uses the back button in Compose UI?
* Can we omit the package name in the path?
* I think it's unnecessary to encode a text after hash (`#`), then `Foo/123` would look nicer
  * Can we turn off encoding on the project site?

cc @terrakok

zsmb13 (approved):
I rebased to latest (because some source sets changed) and renamed it, otherwise looks good!

I have similar notes as @bashor:
* It would be nice to be able to drop the package name at least, seems like that would be safe to do most of the time (can be opt in as well)
* I don't think we can drop suffixes like `Screen`, even in this app we're not super consistent about naming our nav params
* Our Main screen has a nested nav controller (not just nested graph, its own controller as well), I wonder if we can do something to make that work with URLs too

PR #345, opened 2025-03-22
Small UI fixes

zsmb13:


ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/TalkCard.kt
```diff
@@ -413,7 +429,7 @@ internal fun TalkCardPreview() {
                 status = TalkStatus.Live,
                 onSubmitFeedbackWithComment = { e, s -> println("Feedback, emotion + comment: $e, $s") },
                 onSubmitFeedback = { e -> println("Feedback, emotion only: $e") },
-                onClick = { "Clicked session" },
+                onClick = { println("Clicked session") },
```
e5l, line 432: please consider use logging instead

e5l (changes requested):

ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/TalkCard.kt
zsmb13: This is just the code of the Compose preview, it's never used in production

PR #348, opened 2025-03-24
Sync versions of apps, add script to update all versions, add version to About screen

zsmb13:
image

### DIFF
--- a/androidApp/build.gradle.kts
+++ b/androidApp/build.gradle.kts
@@ -33,8 +33,8 @@ android {
         applicationId = "com.jetbrains.kotlinconf"
         targetSdk = libs.versions.android.targetSdk.get().toInt()
         minSdk = libs.versions.android.minSdk.get().toInt()
-        versionCode = 38
-        versionName = "38.0"
+        versionCode = 62
+        versionName = "39.0.0"
 
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
     }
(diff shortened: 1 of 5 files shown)


androidApp/build.gradle.kts
```diff
@@ -33,8 +33,8 @@ android {
         applicationId = "com.jetbrains.kotlinconf"
         targetSdk = libs.versions.android.targetSdk.get().toInt()
         minSdk = libs.versions.android.minSdk.get().toInt()
-        versionCode = 38
-        versionName = "38.0"
+        versionCode = 62
+        versionName = "25.0.0"
```
e5l, line 37: nope, we can't go back in time

iosApp/KotlinConf.xcodeproj/project.pbxproj
```diff
@@ -348,7 +348,7 @@
 					"$(inherited)",
 					"@executable_path/Frameworks",
 				);
-				MARKETING_VERSION = 3.2.2;
+				MARKETING_VERSION = 25.0.0;
```
e5l, line 351: this change will trigger long full app review after submission. I would avoid this for minor updates

e5l (changes requested):

androidApp/build.gradle.kts
zsmb13: This is only the user-visible version string on Android, as far as I know, we can pick any arbitrary value for it. There's nothing checking that it increments over time.

https://developer.android.com/studio/publish/versioning#versioningsettings

iosApp/KotlinConf.xcodeproj/project.pbxproj
zsmb13: Don't we want to have one full app review for the 2025 version anyway?

Do you think we should stick with 3.x.x for the entire 2025 year instead? (In that case, we'd have to change the Android version to that as well.)

e5l: we usually have a short one for the patch updates

e5l (approved):
thanks, looks better

PR #353, opened 2025-03-25
Add background to iOS app icons

zsmb13:
image


e5l (approved):
lgtm

kropp:
@zsmb13 do you remember why the background was needed? It seems that because of that, the icon no longer changes to the dark theme.

PR #360, opened 2025-03-27
Add "Watch the video" button

zsmb13:
image


shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/View.kt
```diff
@@ -93,7 +93,8 @@ fun Session.asSessionCard(
         tags = tags ?: emptyList(),
         startsInMinutes = (startsAt - now).inWholeMinutes.toInt().let { diff ->
             if (diff in 1..30) diff else null
-        }
+        },
+        videoUrl = "https://www.youtube.com/watch?v=Ar73Axsz2YA",
```
e5l, line 97: could you tell me if the link will be static?

zsmb13: Oops! Thanks for catching this, it was my temporary change for testing. It should be mapped from the session coming from the backend (I know we don't send this field yet, this is preparation for later).

Fixed the commit now.

e5l (approved):
lgtm

PR #365, opened 2025-03-27
Remove animations between Main screen destinations

### LINKED ISSUE #364: Bottom navigation tabs should switch without animations on iOS

zsmb13:
Fixes #364 

Removes regular enter/exit animations on these screens, so tapping a bottom nav destination switches instantly.

Adds a basic fade in/out for popping, so that Android predictive back still shows the user where they're going back.

Swiping back to navigate up the stack is disabled on iOS by adding a `BackHandler` that consumes the back gestures.


shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/MainScreen.kt
```diff
@@ -81,8 +88,13 @@ fun MainScreen(
             modifier = Modifier
                 .fillMaxWidth()
                 .weight(1f),
+            enterTransition = { EnterTransition.None },
+            exitTransition = { ExitTransition.None },
+            popEnterTransition = { fadeIn() },
+            popExitTransition = { fadeOut() },
         ) {
             composable<InfoScreen> {
+                MainBackHandler()
```
e5l, line 97: It's weird to do this manually in every screen

e5l (approved):
lgtm

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/MainScreen.kt
zsmb13: Unfortunately the handler has to be inside the nav destination for it to capture back gestures from there. At least we can do it here in the nav graph setup instead of having to put it directly inside each screen.

PR #386, opened 2025-03-31
Improve iOS keyboard handling, add feedback and keyboard behavior flags

zsmb13:
* Fixes scrolling issues caused by the old UIScrollView two-way sync implementation
* Adds a `hideKeyboardOnDrag` flag which, with the bugfix above, should make iOS feedback on the schedule screen work nicely
* Adds a `redirectFeedbackToSessionPage` feature just in case, which replaces in-schedule feedback with navigating the user to the session detail page, to leave their feedback there instead

### Demos

Hide on drag demo:

https://github.com/user-attachments/assets/7bccfc97-c182-43b5-917b-36d83ea3ddfa

Redirect to session for feedback:

https://github.com/user-attachments/assets/4c23083b-fe56-4c7f-b56d-3bd200b5f9be

### DIFF
--- a/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/ScheduleScreen.kt
+++ b/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/ScheduleScreen.kt
@@ -50,6 +50,7 @@ import kotlinconfapp.ui_components.generated.resources.search_24
 import kotlinx.coroutines.launch
 import org.jetbrains.compose.resources.stringResource
 import org.jetbrains.kotlinconf.DayValues
+import org.jetbrains.kotlinconf.LocalFlags
 import org.jetbrains.kotlinconf.ScrollToTopHandler
 import org.jetbrains.kotlinconf.SessionCardView
 import org.jetbrains.kotlinconf.SessionId
@@ -87,6 +88,7 @@ import kotlinconfapp.ui_components.generated.resources.Res as UiRes
 fun ScheduleScreen(
     onSession: (SessionId) -> Unit,
     onPrivacyPolicyNeeded: () -> Unit,
+    onRequestFeedbackWithComment: (SessionId) -> Unit,
     viewModel: ScheduleViewModel = koinViewModel(),
 ) {
     val scope = rememberCoroutineScope()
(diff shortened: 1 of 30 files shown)


shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/SessionScreen.kt
```diff
@@ -126,62 +128,80 @@ fun SessionScreen(
...
+                ScrollToTopHandler(listState)
 
-                    speakers.forEach { speaker ->
-                        SpeakerCard(
-                            name = speaker.name,
-                            title = speaker.position,
-                            photoUrl = speaker.photoUrl,
-                            modifier = Modifier.padding(vertical = 12.dp).fillMaxWidth(),
-                            onClick = { onSpeaker(speaker.id) }
+                LazyColumn(
```
zsmb13, line 141: Swapped this to a LazyColumn so we can scroll precisely to its second item for the feedback block

zsmb13:
Accidentally merged #387 into this branch instead of waiting for this PR to go in first, but it's fine, we can just merge both into the main dev branch from here.

ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/TalkCard.kt
```diff
@@ -377,8 +391,17 @@ private fun FeedbackBlock(
                                 if (userSignedIn) {
```
e5l, line 391: should we extract extension?

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/ScheduleScreen.kt
```diff
@@ -587,6 +610,11 @@ private fun SessionCard(
         onSubmitFeedback = { emotion ->
             onSubmitFeedback(session.id, emotion)
         },
+        onRequestFeedbackWithComment = if (onRequestFeedbackWithComment != null) {
```
e5l, line 613: ?.let would look nice here

e5l (approved):
lgtm, please check comments

ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/TalkCard.kt
zsmb13: I'm not sure which part you mean, but we can do it later :)

PR #393, opened 2025-04-02
Inject app-level CoroutineScope

zsmb13:


kropp:
`CoroutineScope` injection via constructor parameter simplifies the testing significantly!

kropp (approved):

PR #397, opened 2025-04-03
Policy updates

zsmb13:
* Updated *App Privacy Policy* and *App Terms of Use*.
* Added an optional link handler for Markdown, connected these two screens using links.
* Renamed the various policy screens that we have in the app to be more descriptive
     * `AppPrivacyPolicy` and `AppPrivacyPolicyPrompt` are the same document, but the first is read-only, the second is with Accept/Reject options
     * `AppTermsOfUse` is unchanged
     * `VisitorTermsOfUse` is the visitor terms of use doc
     * `VisitorPrivacyPolicy` is the visitor privacy policy doc

https://github.com/user-attachments/assets/d88e23c6-0b8a-430e-99a9-6578e7e28d1b


shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/navigation/Routes.kt
```diff
@@ -35,16 +35,16 @@ data object StartPrivacyPolicyScreen
 data object StartNotificationsScreen
 
 @Serializable
-@SerialName("PrivacyPolicy")
-data object PrivacyPolicyScreen
+@SerialName("AppPrivacyPolicyPrompt")
+data object AppPrivacyPolicyPrompt
```
kropp, line 39: Is `@SerialName` required in this case?

ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/MarkdownView.kt
```diff
@@ -1,73 +1,104 @@
...
 
-    MarkdownView(text, modifier)
+    MarkdownView(text, modifier, onUriClick)
 }
 
 @Composable
 fun MarkdownView(
     text: String,
     modifier: Modifier = Modifier,
+    onUriClick: (String) -> Unit = {},
```
kropp, line 48: The name is a bit misleading, I was about to write a comment, why no URIs are handled by default.
Maybe rename it to something like `onCustomUriClick`, `onInAppUriClick`, or `onLocalUriClick`?

PR #405, opened 2025-04-04
Update session handling logic

zsmb13:
* Skips time slots with no sessions in them immediately when they're being created so we don't have to remove these later
* Improves the "Now" concept with `firstActiveIndex` / `lastActiveIndex`, making sure we point "Now" to the current live or the next upcoming session when the time is somewhere within the conference (even if that's the next day)
* Puts lightning talks at the ends of time slots

### DIFF
--- a/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/View.kt
+++ b/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/View.kt
@@ -72,7 +72,7 @@ fun List<Session>.groupByTime(
                 startsAt = slot.startsAt,
                 endsAt = slot.endsAt,
                 state = SessionState.from(slot.startsAt, slot.endsAt, now),
-                sessions = sessions,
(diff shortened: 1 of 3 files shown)


shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/View.kt
```diff
@@ -42,34 +40,44 @@ fun Conference.buildAgenda(
...
-            sessions = cards,
+    this.forEach { session ->
+        val slot = slots.find { (start, end) -> session.startsAt >= start && session.endsAt <= end } ?: return@forEach
+        slotsToSessions.getValue(slot).add(
+            session.asSessionCard(conference, now, favorites, votes[session.id])
         )
     }
+
+    return slotsToSessions.mapNotNull { (slot, sessions) ->
+        if (sessions.isNotEmpty()) {
```
e5l, line 70: invert if and early return would look nice

e5l (approved):

PR #406, opened 2025-04-04
Add Bloomberg to partners

zsmb13:


shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/HardcodedValues.kt
```diff
@@ -77,6 +78,7 @@ internal val PARTNERS: Map<StringResource, List<Partner>> = mapOf(
         Partner(PartnerId("uber"), "Uber", "We reimagine the way the world moves for the better Movement is what we power.", Res.drawable.partner_uber, "https://www.uber.com/nl/en/"),
     ),
     Res.string.partners_bronze to listOf(
+        Partner(PartnerId("bloomberg"), "Bloomberg", "", Res.drawable.partner_bloomberg, "https://www.bloomberg.com/company/values/tech-at-bloomberg/"),
```
zsmb13, line 81: We don't have a description for now, it's intentionally missing

e5l (approved):

PR #407, opened 2025-04-06
Fix three-button nav bar

Goooler:
| Before | After |
|--------|--------|
| before | after |


zsmb13:
On a device set to dark theme with a custom light theme selected in the app, this leads to the scrim being removed, and the nav buttons being effectively invisible, white on white:

Screenshot_20250406-150331

zsmb13:
On the other hand, the current theming looks like this for me, with the app forced to dark mode, and the system set to light/dark:

Goooler:
Everything should be all right now.

https://github.com/user-attachments/assets/beffaf53-87c1-4712-a6cd-32accecc97ae

PR #417, opened 2025-04-08
Accessibility improvements

eevajonnapanula:
First patch of accessibility improvements. They consist of the following:
- Adding a couple of content descriptions that were missing/incorrectly set.
- Adding state descriptions for component benefitting from them.
- Annotating headers for easier navigation for screen reader (and other assistive tech utilizing them) users
- Marking errors as live regions so that they'd be announced
- Changing some `clickable` modifiers to `selectable` or `toggleable` to match the semantics to the usage of the element
- Fixing some issues with large text (see example below)

In general, the other changes are about the semantics changing + some tab stops disappearing, the only UI change is the privacy policy screen's scrollability:

<table>
<tr>
  <td>Before</td>
  <td>After</td>
</tr>
<tr>
  <td><video src="https://github.com/user-attachments/assets/3002c39c-971f-4a24-898d-98f349791c73" controls /></td>
  <td><video src="https://github.com/user-attachments/assets/851b7cb3-43ce-49f6-a8dc-b4a4fe0fddc2" controls /></td>
</tr>
</table>

### DIFF
--- a/ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/ErrorLoading.kt
+++ b/ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/ErrorLoading.kt
@@ -38,6 +38,9 @@ import androidx.compose.ui.geometry.Offset
 import androidx.compose.ui.geometry.Size
 import androidx.compose.ui.graphics.lerp
 import androidx.compose.ui.graphics.vector.rememberVectorPainter
+import androidx.compose.ui.semantics.LiveRegionMode
+import androidx.compose.ui.semantics.liveRegion
+import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.dp
 import kotlinconfapp.ui_components.generated.resources.Res
@@ -64,7 +67,15 @@ fun MinorError(
     message: String,
     modifier: Modifier = Modifier,
 ) {
-    ErrorText(message, modifier.fillMaxWidth().padding(32.dp))
+    ErrorText(
+        message,
+        modifier
+            .fillMaxWidth()
+            .padding(32.dp)
+            .semantics {
+                liveRegion = LiveRegionMode.Assertive
+            }
+    )
 }
 
 @Composable
@@ -82,7 +93,12 @@ fun MajorError(
             contentDescription = null,
         )
         Spacer(Modifier.height(16.dp))
-        ErrorText(message)
+        ErrorText(
+            message = message,
+            modifier = Modifier.semantics {
+                liveRegion = LiveRegionMode.Assertive
+            }
+        )
     }
 }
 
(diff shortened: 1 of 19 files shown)


ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/ErrorLoading.kt
```diff
@@ -64,7 +67,15 @@ fun MinorError(
...
     modifier: Modifier = Modifier,
 ) {
-    ErrorText(message, modifier.fillMaxWidth().padding(32.dp))
+    ErrorText(
+        message,
+        modifier
+            .fillMaxWidth()
+            .padding(32.dp)
+            .semantics {
+                liveRegion = LiveRegionMode.Assertive
```
zsmb13, line 76: Wow, TIL!

ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/Switcher.kt
```diff
@@ -56,7 +58,11 @@ private fun SwitcherItem(
                 color = strokeColor,
                 shape = SwitcherItemShape,
             )
-            .clickable(onClick = onClick)
+            .toggleable(
```
zsmb13, line 61: As these can't be un-toggled,  should they be a `selectable` instead?

ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/Toggle.kt
```diff
@@ -66,11 +66,6 @@ fun Toggle(
         Box(
             Modifier.size(ToggleWidth, ToggleHeight)
                 .clip(ToggleBackgroundShape)
-                .clickable(
-                    onClick = { onToggle(!enabled) },
-                    interactionSource = interactionSource,
-                    indication = LocalIndication.current,
-                )
```
zsmb13, line 73: These were just redundant, but not incorrect, right? Having them set here gives the toggle on its own a ripple when clicked (directly) that's why they were used.

zsmb13 (approved):
This looks great, thank you for contributing!

I pushed a couple commits, as I realized we were missing a title on the settings screen as I was going through the changes, plus I updated some string resources.

Left a couple comments.

ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/Toggle.kt
eevajonnapanula: This is actually something I was thinking a lot. From assistive tech-navigation-POV, they do add extra tab stops - I'm wondering, is there a way to make a component non-focusable but clickable, because that would be the best option here 🤔

ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/Switcher.kt
eevajonnapanula: Yeah, true, now that I'm thinking, `selectable` is definitely the right choice here - as these are `Tab`s, and can be only selected one at a time. Will change that one (+ add the `selectableGroup` to add list semantics)

ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/Toggle.kt
zsmb13: I guess the weird part is that we don't use this directly (so far), only embedded into `SettingsItem`, where we clear its semantics so that we can put the listener on the larger piece of UI (the whole row) instead and not just the toggle itself.

But in theory this toggle could be used on its own as well somewhere else.

PR #423, opened 2025-04-09
Fix padding on speaker detail screen

zsmb13:
Fix the issue where the padding was previously applied outside of the scrollable area. The content now correctly reaches the top toolbar when scrolled:

https://github.com/user-attachments/assets/7b740bac-7bac-4aea-9bcc-2f9abde7c872

Modifier ordering is hard


emartynov:
LGTM.

Off topic question - no snapshot tests?

zsmb13:
@emartynov We're in a no tests stage at the moment, but we should definitely have them for a bunch of things when we find the time.

kropp (approved):

PR #425, opened 2025-04-09
Improve search and filters

zsmb13:
* Move filter into scrollable list
* Show result count on both search screens

https://github.com/user-attachments/assets/466937b4-d784-48d0-8d1f-38f47f550ee1

https://github.com/user-attachments/assets/df6bc73c-239a-4d26-8aa1-66efcaef8c80


e5l (approved):
lgtm

PR #432, opened 2025-04-12
Add in-memory debug logging flag to dev menu

zsmb13:
image


shared/src/jvmMain/kotlin/org/jetbrains/kotlinconf/main.kt
```diff
@@ -27,8 +25,14 @@ private val platformModule = module {
     }
 }
 
+class JvmLogger : Logger {
+    override fun log(tag: String, lazyMessage: () -> String) {
+        println("[$tag] ${lazyMessage()}")
```
e5l, line 30: it would be nice to use logging library

e5l (approved):

PR #434, opened 2025-04-12
Only request focus once in feedback fields

### LINKED ISSUE #430: Focus requesters on the schedule screen re-request focus too many times
We should only request focus for the text field when it initially appears as the talk card expands, but currently if we just scroll away and then scroll back to the given session, it'll steal focus (because the expanded state is saveable and gets restored, but we don't remember in any way whether we already requested focus there)

zsmb13:
Fixes #430

### DIFF
--- a/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/SessionScreen.kt
+++ b/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/SessionScreen.kt
@@ -347,9 +347,13 @@ private fun FeedbackPanel(
         }
 
         AnimatedVisibility(visible = feedbackExpanded) {
+            var focusRequested by rememberSaveable { mutableStateOf(false) }
             val focusRequester = remember { FocusRequester() }
-            LaunchedEffect(Unit) {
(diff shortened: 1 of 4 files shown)


shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/SessionScreen.kt
```diff
@@ -347,9 +347,13 @@ private fun FeedbackPanel(
         }
 
         AnimatedVisibility(visible = feedbackExpanded) {
+            var focusRequested by rememberSaveable { mutableStateOf(false) }
```
e5l, line 350: I would keep the state of request inside focusRequester

e5l (approved):

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/SessionScreen.kt
zsmb13: `FocusRequester` is a framework class, so it won't track this sort of thing for us

PR #436, opened 2025-04-14
Add Junie badge

zsmb13:
image
image


shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/AboutAppScreen.kt
```diff
@@ -98,6 +107,12 @@ fun AboutAppScreen(
                     }
                     .padding(16.dp)
             )
+
+            Image(
+                imageVector = vectorResource(Res.drawable.made_with_junie),
+                contentDescription = stringResource(Res.string.about_app_made_with_junie),
+                modifier = Modifier.align(Alignment.CenterHorizontally)
+            )
```
kropp, line 115: Let's make it clickable and open landing page

kropp (changes requested):

shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/screens/AboutAppScreen.kt
zsmb13: image

kropp (approved):

PR #437, opened 2025-04-15
A11y: Carousel semantics

eevajonnapanula:
This PR adds some more semantics (headings & merges service items into one block), and adds correct semantics for the workshop carousel. 

For some reason, the `HorizontalPager` and its child items don't have the correct list semantics, so I added `collectionItemInfo` and `collectionInfo` to them. And even if everything worked as expected with `HorizontalPager`, there's a problem with the page count; now it has `Int.MAX_SIZE` as the size of the list of items (`pageCount`), so instead of the seven items the list has, `Int.MAX_SIZE` would be announced. These changes fix that as well.


zsmb13 (approved):

zsmb13:
Thanks again for the contribution! 🙌

PR #440, opened 2025-04-15
Update policies

zsmb13:
https://github.com/user-attachments/assets/5b0a260d-2f53-4c16-a237-3678492ee177


e5l (approved):
should we request them again after updating?

zsmb13:
@e5l Thankfully it's the visitor policies that changed in their content and not the app policies, but maybe this is something we'll need to implement later

PR #453, opened 2025-04-22
Fix first scroll logic

zsmb13:
When not in search, scroll to active index on first load.

When in search, scroll to top when entering search or changing the search query.

https://github.com/user-attachments/assets/ae6a2afd-7d54-44f4-8334-e216c01a13ea


kropp:
It is unclear to me, why do we scroll to top when the search field is opened but still empty?
Wouldn't it be better if the content stays where it is until the first character is typed?

kropp (approved):

PR #462, opened 2025-04-24
update Compose Multiplatform to 1.8.0-rc01

kropp:
`navigation-compose` no longer exposes `lifecycle-common-compose` as `api` dependency (rightfully so!), need to add it explicitly


zsmb13 (approved):
rc01 let's go 🚀 Good to merge as soon as CI is green

PR #466, opened 2025-04-29
Fix potential crash by replacing first() with firstOrNull() in FlagsManager initialization

ChoiSeungMyung:
### Background
In `FlagsManager`, during initialization, the code currently retrieves the stored flags by calling `storage.getFlags().first()`.  
If the `Flow` is empty (i.e., no flags have been stored yet), `first()` throws a `NoSuchElementException`, which can crash the application.

Given that the code checks whether `storedFlags` is `null`, it is more appropriate and safer to use `firstOrNull()` instead.  
This way, if the `Flow` does not emit any value, it returns `null` instead of crashing, allowing the initialization logic to proceed correctly.

Additionally, since the `flags` property already handles fallback to `platformFlags` when necessary (via `map { it ?: platformFlags }` and `stateIn`),  
this change ensures consistency and robustness without altering the intended behavior.

---

### Changes
- Replace `first()` with `firstOrNull()` when retrieving stored flags in `FlagsManager.init`.

---

### Why
- Prevents potential `NoSuchElementException` and application crashes when no flags have been stored.
- Aligns with the existing `null` check (`storedFlags == null`) in the initialization logic.
- Improves overall code robustness and consistency.

---

### DIFF
--- a/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/FlagsManager.kt
+++ b/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/FlagsManager.kt
@@ -2,7 +2,7 @@ package org.jetbrains.kotlinconf
 
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.SharingStarted
-import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.firstOrNull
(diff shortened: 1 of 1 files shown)


zsmb13:
`first` only throws if the `Flow` was completely empty and completed, which can not happen in this case.

PR #477, opened 2025-05-07
Wrap Icon in TopMenuButtonImpl in a TooltipBox to be able to show a tooltip for Toolbar menu items

### LINKED ISSUE #476: Missing tooltips for Toolbar menu items
On Android, when a user long presses on a Toolbar menu item, a tooltip shold show the name of the option. The KotlinConf app does not do this while it would be a nicer user experience to do so, removing any disambiguity from what icons mean, which might not be obvious for all users.

balazsgerlei:
Fixes #476 

- On Android, long pressing on a Toolbar menu item should display the menu items' names in a tooltip
- The Tooltip's background color is not correct yet in light mode (it uses a dark color for some reason)
- Content description is used as the tooltip text

### DIFF
--- a/ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/TopMenuButton.kt
+++ b/ui-components/src/commonMain/kotlin/org/jetbrains/kotlinconf/ui/components/TopMenuButton.kt
@@ -8,7 +8,12 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.selection.toggleable
 import androidx.compose.foundation.shape.CircleShape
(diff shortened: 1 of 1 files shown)


balazsgerlei:
As I wrote, unfortunately the background color of the tooltip is not correct in light theme, it still uses a dark color. OFC it could be solved with specifying the `containerColor` in `PlainTooltip`, but obviously that should not be the answer.

Probably something in themes should be modified. Can you help with this @zsmb13? Thanks!

zsmb13:
Since we removed Material 3 in #553, I've added a commit that drops down to using the Foundation APIs for tooltips instead. These seem reasonably good too.

For colors and design, I've temporarily set it to some existing color tokens from our design system - but I'll ask for input from the app's designer on what these should look like and get back to this.

PR #483, opened 2025-05-12
Remove `KmpNotifier` library export from iOS source set

mirzemehdi:
This PR removes the export of the kmpnotifier library in the iOS framework generation.

`IOSNotifications.kt` file in the `iosMain` source set already handles iOS notifications using Kotlin. Exporting `kmpnotifier` library for use on the Swift side is unnecessary. All notification-related logic can be managed within the kotlin side.


zsmb13 (approved):

zsmb13:
Thanks for the contribution! 🙌

PR #511, opened 2025-07-03
Sessionize video url

zsmb13:


zsmb13:
@e5l In progress here https://github.com/JetBrains/kotlinconf-app/pull/512

PR #531, opened 2025-08-07
Backend: remove GlobalScope usage

nomisRev:
Refactor `NewsService` and `SessionizeService` to use `CoroutineScope` for job lifecycle management and adjust DI module accordingly.

Using `GlobalScope` is bad practice, and Ktor already provides a `CoroutineScope` that provides structured concurrency behavior according to the `Application` lifecycle through `Application` itself.

This PR removes the `GlobalScope` usage, and replaces it by an injected `CoroutineScope` which is the encapsulated `Application`.

### DIFF
--- a/.github/workflows/build-all.yml
+++ b/.github/workflows/build-all.yml
@@ -50,7 +50,7 @@ jobs:
           distribution: 'zulu'
           java-version: 21
       - name: Set Xcode version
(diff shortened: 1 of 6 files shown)


zsmb13 (approved):
LGTM but adding @e5l to have a look as well since it's the backend :)

PR #551, opened 2025-11-05
Create separate model module

zsmb13:
Creates a new `core` module that `shared` and `backend` depend on. This way the backend no longer depends on all the client-side implementation in `shared` and `ui-components`, or the Compose framework.


zsmb13:
@e5l This name is the same as the currently proposed default name for the shared module between client/server in the new project structure.

PR #699, opened 2026-04-22
Koin Annotations 4.2.1 + Koin Compiler Plugin 1.0.0-RC2

arnaudgiuliani:
## Proposal: migrate DI to Koin Annotations 4.2.1 + Koin Compiler Plugin 1.0.0-RC2

Hello JetBrains team 👋 

The KotlinConf app originally ran on the **Koin DSL**, then moved to **Metro** to gain compile-time graph verification. This PR proposes a third step: **Koin Annotations 4.2.1** with the new **Koin Compiler Plugin 1.0.0-RC2**.

The goal is to get back to the Koin you already know, familiar APIs, `startKoin { ... }`, `koinInject`, `koinViewModel`, but with an annotations-based declaration and **compile-time safety on par with Metro**. No more trade-off between "ergonomic Koin" and "graph verified at build time."

**Net result across the shared module + 4 platform apps:** `80 files changed, 372 insertions(+), 633 deletions(-)`: about 40% less DI code than the Metro version, with zero change in runtime behaviour.

### 1. A single `@KoinApplication` replaces the hand-written graph interface

The old `AppGraph` (133 lines) had to enumerate every service exposed to the app, declare `@Provides`/`@SingleIn(AppScope::class)` on each factory, wire a custom `MetroViewModelFactory`, and expose a `Map<KClass<out ViewModel>, Provider<ViewModel>>` for assisted ViewModels. It's now:

```kotlin
@KoinApplication
class KotlinConfKoinApp

@Module
@ComponentScan("org.jetbrains.kotlinconf")
@Configuration
class AppModule {
    @Singleton fun provideAppScope(): CoroutineScope = ...
    @BaseUrl @Singleton fun provideBaseUrl(...)
    @Singleton fun provideHttpClient(...)
    @Singleton fun provideTimeProvider(...)
}
```

No more `ViewModelGraph`, no more exposing every service as a property on the graph, no more `LocalAppGraph` / `LocalMetroViewModelFactory` composition locals. `App()`, `MainActivity`, and every platform `main()` stopped taking an `AppGraph` parameter. Composables use `koinInject<ConferenceService>()` directly, exactly like the original DSL-era code.

### 2. ViewModels lose their factory boilerplate

Every ViewModel with a runtime parameter used to carry ~8 lines of factory scaffolding:

```kotlin
// Before (Metro)
@AssistedInject
class SpeakerDetailViewModel(
    private val service: ConferenceService,
    @Assisted speakerId: SpeakerId,
) : ViewModel() {
    ...
    @AssistedFactory
    @ManualViewModelAssistedFactoryKey
    @ContributesIntoMap(AppScope::class)
    fun interface Factory : ManualViewModelAssistedFactory {
        fun create(speakerId: SpeakerId): SpeakerDetailViewModel
    }
}
```

```kotlin
// After (Koin Annotations)
@KoinViewModel
class SpeakerDetailViewModel(
    private val service: ConferenceService,
    @InjectedParam speakerId: SpeakerId,
) : ViewModel() { ... }
```

Call-site becomes `koinViewModel<SpeakerDetailViewModel> { parametersOf(speakerId) }`; same ergonomics as the DSL era, no factory interface to maintain. The pattern repeats across 10+ ViewModels.

### 3. Platform modules collapse to plain classes

`BaseAndroidAppGraph`, `JvmAppGraph`, `IosAppGraph`, `WebAppGraph` were interfaces extending `AppGraph` with `@Provides @SingleIn(AppScope::class)` on each method. They become plain `@Module @Configuration` classes with `@Singleton` functions; `JvmAppGraph` (45 lines with a `@DependencyGraph.Factory` nested interface) is now `BaseJvmModule` (34 lines, no factory needed).

### 4. Compile-time safety — the point of this migration

This is the key argument for staying with Koin rather than a different DI framework: **Koin Compiler Plugin 1.0.0-RC2 validates the DI graph at build time**. Missing bindings, unresolved qualifiers (`@BaseUrl`, `@FileStorageDir`, `@Year`), scope mismatches. All fail `./gradlew build`, just like they did with Metro. You get the same safety net you moved to Metro for, without the graph-interface overhead and without leaving the Koin ecosystem.

### 5. Platform entry points simplify

```kotlin
// Desktop — before
val graph = createGraphFactory<JvmAppGraph.Factory>()
    .create(platformFlags = Flags(supportsNotifications = false, debugLogging = true))
initApp(appGraph = graph, platformLogger = JvmLogger())
...
App(graph)

// Desktop — after
initApp(
    platformLogger = JvmLogger(),
    platformFlags = Flags(supportsNotifications = false, debugLogging = true),
)
...
App()
```

Android / iOS / Web follow the same pattern. Platform-specific config (Android `NOTIFICATION_ICON`, iOS `NotificationPlatformConfiguration`, desktop flags) is passed through the standard `startKoin { androidContext(...); properties(...) }` block. Exactly the Koin API the app started with.

### Scope (`YearScope`)

The per-year scope (swapping `YearlyStorage` + `YearlyApi` when the server pushes a new conference year) is preserved via Koin's `@Scope(YearScope::class)` + `KoinScopeComponent`. Same semantics as the old `@GraphExtension(YearScope::class)`.

### Versions bumped

```
koin = "4.2.1"
koin-plugin = "1.0.0-RC2"   # Koin Compiler Plugin (new — provides build-time verification)
```

Note: `1.0.0-RC2` is a release candidate; GA is imminent. Happy to split this into smaller commits per platform if that's easier to review, and to walk through any specific ViewModel or scope case on a call.

— 

Your feedback is welcome :)


zsmb13:
Hi @arnaudgiuliani, thanks for the PR! I think it's very valuable to have as a real-life comparison of the frameworks if anyone’s looking for one.

**TL;DR from my perspective:** I don’t see a compelling reason to transition from Metro to Koin in this project.

- It’s clear that Koin _can_ work for a project like this, and that it has some implicit behaviors and design choices that make its DI configuration somewhat simpler in terms of the annotations and lines of code required.
- However, these don’t change the fundamental experience of working with DI in the project, and there _are_ certain safety-related things that don’t seem to be on par with what Metro provides.

---

Below are my findings in more specific detail:

- Assisted injection is certainly a bit simpler, as it doesn’t require a Factory interface to be defined. The call site is also a little simpler, as it avoids some generic type parameters.
   - But the API there (`parametersOf`) doesn’t seem to offer compile time safety. If I provide extra parameters, they are silently ignored. If I forget to provide a parameter, it compiles successfully and fails at runtime.
- If I introduce a circular dependency, that still compiles with no errors, and crashes only at runtime.
- Grabbing dependencies from the DI container is also easier, as you can just `koinInject` anything anywhere, which grabs values from the static global instance (vs having to expose these explicitly on the app graph and having to provide access to the app graph somehow, like we do with composition locals in our Metro setup).
   - It looks like these _are_ verified, and I get a compilation error if the type I’m trying to inject is not available in the container.
   - On the other hand, being able to do this anywhere and grab any type from the container might also be undesirable, especially in large projects with lots of objects available as dependencies.
- Similarly, it slightly concerns me that the Koin instance is globally accessible *and* mutable at runtime, allowing code to arbitrarily modify the contents, declaring new dependencies, editing scopes, and so on.
- We get to skip the `Factory` interfaces for each platform’s specific graph.
- ViewModels are scoped globally implicitly, so no need to specify that scope for each, saving an annotation. There are also other similar annotations (usually, `@Inject`) we can skip as they are implicit with Koin.
- ViewModel and Android app component support is included, which skips some one-time setup (like the Metro ViewModelFactory implementation).
- The changes around `YearGraph` / `YearScope` aren’t super clear for me, but they seem functionally correct.
- Since Koin has a global static instance, we’re saving the couple lines of code at each app initialization that would create the DI graph and store it in a property.
- I don’t particularly understand what’s happening in `initApp` and why it’s happening, compared to was there before.

---

The overall reduction is ~240 lines of code, on our codebase of ~20k lines.

- ~100 of this is simply fewer imports
- ~60 is saving annotations where some things are implicit in Koin but explicit in Metro
- ~40 is the assisted factories for VMs

---

I don't have comments on performance, I haven't measured anything at either compile time or runtime.

arnaudgiuliani:
Thanks @zsmb13 for the detailed review. And totally fair call not to switch DI this close to the KotlinConf event.

Two fair points on compile safety, both already on the 1.0 milestone:

- parametersOf lacks compile-time validation: correct today. The compileSafety pass in the compiler plugin already walks call sites for missing-dependency checks, and extending it to validate parametersOf argument count and types is an incremental addition on that same pass.

- Circular dependencies crash at runtime: also correct, and a direct extension of the same validation. The dependency graph is already reconstructed at compile time for safety checks; cycle detection is a graph-walk on top.

Compile-safety is already working well today (modules and call-site validation on koinInject, koinViewModel, etc.), and both gaps above are direct extensions of that same pipeline.

---

On the other points, a few clarifications in context:

- On DI fundamentals. The experience may look similar (constructor injection, modules...), but there is a meaningful architectural difference. With a graph-first approach like Metro, every feature module needs explicit access (factory interfaces per platform, composition locals to expose it). That plumbing scales with the surface area: each new feature adds new coordination points across the codebase.

   With Koin, the container is more ambient (in its default mode), so that coordination layer largely disappears. Feature teams declare what they need, and the container resolves it. In practice, this shifts complexity away from the application layer and avoids the need to continuously evolve a shared graph structure as the app grows.

- On koinInject. Visibility rules (internal, private, module boundaries) combined with compile-time call-site validation make it scoped and verified, not “anything goes”. The main difference is where scoping lives (explicit graph topology vs module boundaries + ambient access), and how quickly teams can evolve. In practice, koinInject tends to be faster to adopt for feature teams, since adding a dependency doesn’t require extending a shared graph schema.

- On reachability. Metro makes it explicit through graph/subgraph structure; Koin partitions it through scopes and modules, layered with Kotlin visibility: a definition that isn’t visible can’t be injected. @Configuration adds another lever for dynamic discovery while keeping definitions scoped where they belong. Same architectural discipline applies in both models, with different mechanisms.

- On runtime mutability. Koin's runtime mutability is a deliberate choice. It's what enables runtime overrides for feature flags, test modules, and dynamic configuration: workflows that large KMP projects actually rely on. For teams that want compile-time determinism, strict mode rejects re-declarations. Metro picks one answer at compile time; Koin lets you pick per project.

---

Zooming out, this isn’t really about one approach being “better” in the abstract. Both can reach similar safety guarantees. The difference is how they scale in practice. 

For modularized multiplatform apps like KotlinConf, we’ve consistently seen the ambient container model reduce coordination overhead between feature modules, avoid graph maintenance costs, and keep development velocity high as the codebase grows. That’s ultimately why we think Koin is a strong fit for this kind of architecture.

---

On initApp and YearScope, I’ll add a follow-up comment with the diff vs the prior setup. YearScope was kept close to your patterns to expose fields and avoid breaking changes.

The Koin team will keep maintaining the KotlinConf app with the Koin version on its official repo (https://github.com/InsertKoinIO/kotlinconf-app). 

Thanks again for the feedback ✌️

